PR #49, opened 2022-10-18
Support InterstateChannel command

longquanzheng:


duoertai (commented):
so the InterStateChannelPublishing in StateStartResponse is for telling iwf value and channel to publish to, and the InterStateChannelCmd and CmdResult is telling iwf what channel to expect for a value/publish, is that right?

duoertai (approved):

PR #60, opened 2022-11-01
add getSearchAttribute

duoertai:
temporal tested
cadence tested

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "iwf-idl"]
 	path = iwf-idl
-	url = git@github.com:cadence-oss/iwf-idl.git
+	url = git@github.com:indeedeng/iwf-idl.git
--- a/service/api/temporal/client.go
+++ b/service/api/temporal/client.go
@@ -115,15 +115,15 @@ func mapToIwfSearchAttributes(searchAttributes *common.SearchAttributes) (map[st
 		if isString {
 			result[key] = iwfidl.SearchAttribute{
 				Key:         iwfidl.PtrString(key),
-				StringValue: &str,
+				StringValue: iwfidl.PtrString(str),
 				ValueType:   iwfidl.PtrString(service.SearchAttributeValueTypeKeyword),
 			}
 		}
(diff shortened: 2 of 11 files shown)


service/api/temporal/client.go
```diff
@@ -82,16 +85,59 @@ func (t *temporalClient) DescribeWorkflowExecution(ctx context.Context, workflow
...
+
+		str, isString := object.(string)
+		if isString {
+			result[key] = iwfidl.SearchAttribute{
+				Key:         iwfidl.PtrString(key),
+				StringValue: &str,
+				ValueType:   iwfidl.PtrString(service.SearchAttributeValueTypeKeyword),
+			}
+		}
+		integer, isInt := object.(float64)
```
duoertai, line 122: @longquanzheng debugged and found the returned type is float64..., maybe a bug in temporal

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "iwf-idl"]
 	path = iwf-idl
-	url = git@github.com:cadence-oss/iwf-idl.git
+	url = git@github.com:indeedeng/iwf-idl.git
```
duoertai, line 3: updated this because it's blocking updating idl

longquanzheng: I think you also need to commit the commitId of the module

service/api/temporal/client.go
longquanzheng: https://github.com/temporalio/sdk-go/issues/942 opened a bug ticket

.gitmodules
duoertai: there is update of iwf-idl submodule change in this PR, but I'm not sure why it's not showing the original commit id and the new commit id, instead it's showing the changes in the file. And if you click into the change, both commit are in the same repo

is it because submodule url is changed?

service/api/temporal/client.go
duoertai: should we do the casting to json.Number/float64 and then get/cast to int64? or should we create a data converter?

longquanzheng (approved):

PR #79, opened 2022-11-14
add cron schedule and workflow id reuse policy

duoertai:


service/api/cadence/client.go
```diff
@@ -143,6 +149,21 @@ func mapToIwfSearchAttributes(searchAttributes *shared.SearchAttributes) (map[st
...
+	case service.WorkflowIDReusePolicyAllowDuplicate:
+		return client.WorkflowIDReusePolicyAllowDuplicate, nil
+	case service.WorkflowIDReusePolicyAllowDuplicateFailedOnly:
+		return client.WorkflowIDReusePolicyAllowDuplicateFailedOnly, nil
+	case service.WorkflowIDReusePolicyRejectDuplicate:
+		return client.WorkflowIDReusePolicyRejectDuplicate, nil
+	case service.WorkflowIDReusePolicyTerminateIfRunning:
+		return client.WorkflowIDReusePolicyTerminateIfRunning, nil
+	default:
+		return client.WorkflowIDReusePolicyAllowDuplicate, fmt.Errorf("unsupported workflow id reuse policy %s", workflowIdReusePolicy)
```
duoertai, line 163: cadence doesn't have a UNSPECIFIED enum

service/api/cadence/client.go
```diff
@@ -36,10 +36,16 @@ func (t *cadenceClient) Close() {
...
+	workflowIdReusePolicy, err := mapToTemporalWorkflowIdReusePolicy(options.WorkflowIDReusePolicy)
+	if err != nil {
+		return "", nil
+	}
 	workflowOptions := client.StartWorkflowOptions{
 		ID:                           options.ID,
 		TaskList:                     options.TaskQueue,
 		ExecutionStartToCloseTimeout: options.WorkflowRunTimeout,
+		WorkflowIDReusePolicy:        workflowIdReusePolicy,
+		CronSchedule:                 options.CronSchedule,
```
duoertai, line 48: @longquanzheng do we need validate cron schedule here?

longquanzheng: I think the Cadence/Temporal go SDK will do it so we don't need to do here

service/api/cadence/client.go
```diff
@@ -143,6 +149,21 @@ func mapToIwfSearchAttributes(searchAttributes *shared.SearchAttributes) (map[st
 	return result, nil
 }
 
+func mapToTemporalWorkflowIdReusePolicy(workflowIdReusePolicy string) (client.WorkflowIDReusePolicy, error) {
```
longquanzheng, line 152: nit: `mapToCadenceWorkflowIdReusePolicy`

service/api/cadence/client.go
longquanzheng: yeah returnin gerror should be fine

longquanzheng: we probably should return nil for the policy (need to change the signature to use pointer)

PR #298, opened 2023-07-06
Update the readme with readability and formatting improvements. 

Bostwickenator:


README.md
```diff
@@ -29,122 +29,122 @@ Related projects:
...
+The `waitUntil` API returns "commands" to wait for. When the commands are completed, the `execute` API will be invoked.
+Both the `waitUntil` and `execute` APIs have access to read/write the persistence schema defined in the workflow.
 
-The full detailed execution flow is like this:
+The full execution flow looks like this:
 
 ![Workflow State diagram](https://user-images.githubusercontent.com/4523955/234921554-587d8ad4-84f5-4987-b838-959869293465.png)
 
-The `waitUntil` API is optional. If not defined, then `execute` API will be invoked instead when the state started.
+The `waitUntil` API is optional. If not defined, then the `execute` API will be invoked immediately when the Workflow State is started.
```
Bostwickenator, line 112: I've changed this to Workflow State. But I don't know if that's clearer or less clear. Internally I guess they are called just states. I could go with capitalized State here if that's more correct.

README.md
```diff
@@ -29,141 +29,138 @@ Related projects:
...
-
-In this case, you need to utilize the "InternalChannel". WorkflowState 4 should be waiting on an "InternalChannel" for 3 messages in "waitUntil" API.
-And then WorkflowState 1,2,3 will be publishing one message for each when completing.  
+In this case, you need to utilize the "InternalChannel". WorkflowState 4 should be waiting on an "InternalChannel" for 3 messages via the `waitUntil` API. 
+WorkflowState 1,2,3 will each publish a message when completing. This ensures propper ordering.  
 
 ## RPC
 
-RPC stands for "Remote Procedure Call". It's for external system to interact with the workflow execution.
+RPC stands for "Remote Procedure Call". Allows external systems to interact with the workflow execution.
```
Bostwickenator, line 163: workflow execution here might be confusingly close to workflow state. It doesn't mean workflow state does it?

README.md
longquanzheng: Definitely better and clearer! Sometimes it’s called “state” just for short, esp in verbal communication . But it’s better to use the full name here.

README.md
longquanzheng: No it’s more than wf states. An RPC and read and mutate the persistence data, publish messages to internal channels and also trigger state execution. Persistence and internal channels are defined on the workflow level. Workflow states are part of the workflow definition

longquanzheng (approved):

PR #300, opened 2023-07-06
More readme updates.

Bostwickenator:
More updates to the readme file. Work in process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Note:
 * With caching enabled read-after-write access will become *eventually consistent*, unless `bypassCachingForStrongConsistency=true` is set in RPC options
 * Caching will introduce an extra event in history (upsertMemo operation for WorkflowPropertiesModified event) for updating the persisted data attributes
 * Caching will be more useful for read-only RPC (no persistence.SetXXX API or communication API calls in RPC implementation) or GetDataAttributes API.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -278,7 +274,7 @@ iWF allows you to start a workflow with a fixed cron schedule like below
 // or timed out, the workflow will be retried based on the retry policy. While the workflow is retrying, it won't
 // schedule its next run. If the next schedule is due while the workflow is running (or retrying), then it will skip
 that
-// schedule. Cron workflow will not stop until it is terminated or cancelled (by returning cadence.CanceledError).
+// schedule. Cron workflow will not stop until it is terminated or canceled (by returning cadence.CanceledError).
```
Bostwickenator, line 277: cadence.CanceledError seems incorrect here given the two backends.

Bostwickenator:
I'll resolve the conflicts shortly.

README.md
longquanzheng: Ahh right. I think we just remove this part. Returning cancel error is not supported in IWF. The only to cancel a workflow is call the stop workflow api ( the stopType is cancel by default)

longquanzheng:
> I'll resolve the conflicts shortly.

Btw there is only one conflict. There is a resolve conflict button in this PR and you can use it to resolve it easily

PR #319, opened 2023-08-03
Support None to not load any data for the persistence loading policy - part 1

zklgame:
Support the None loading type for waitUntil and execute APIs.


integ/util.go
```diff
@@ -196,3 +196,17 @@ func minimumContinueAsNewConfig() *iwfidl.WorkflowConfig {
 		ContinueAsNewThreshold: iwfidl.PtrInt32(1),
 	}
 }
+
+func getBackendTypes() []service.BackendType {
```
longquanzheng, line 200: This is a good idea to improve the code reuse. 
Let's return `[]service.BackendType{ service.BackendTypeTemporal, service.BackendTypeCadence}` if not specify for Temporal or Cadence. This is for making it easy for local test when you click the button to run the test directly. It doesn't supply the flag of cadence or temporal, but we want to do both if possible. 
https://github.com/indeedeng/iwf/blob/main/integ/flag.go#L11

integ/persistence_loading_policy_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	assertions.Equalf(map[string]int64{
+		"S1_start":  1,
+		"S1_decide": 1,
+		"S2_start":  1,
+		"S2_decide": 1,
+	}, history, "persistence loading policy test fail, %v", history)
+
+	if loadingType == persistence_loading_policy.LoadingTypeAll {
+		assertions.Equalf(map[string]interface{}{
```
longquanzheng, line 102: Using the `data` for assertion is fine, but I just want to let you know that there is another way -- you can just do some assertion in the workflow code directly, and panic if not matched, like this one: https://github.com/indeedeng/iwf/blob/dbf26303ecfa3cc8a3dac0ad9f2759a892d98fff/integ/workflow/rpc/routers.go#L232

Initially I always use `data` to assert in test code, but later on I found that putting the assertion in workflow code(routers.go) is simpler to write the test and easier to read as well. :D

integ/persistence_loading_policy_test.go
```diff
@@ -0,0 +1,155 @@
...
+				Encoding: iwfidl.PtrString("json"),
+				Data:     iwfidl.PtrString("test-data-object-value1"),
+			},
+			"sa_state_decide_S2_da_2": iwfidl.EncodedObject{
+				Encoding: iwfidl.PtrString("json"),
+				Data:     iwfidl.PtrString("test-data-object-value2"),
+			},
+		}, data, "persistence loading policy test fail, %v", data)
+	} else if loadingType == persistence_loading_policy.LoadingTypePartialWithoutLock {
+		assertions.Equalf(map[string]interface{}{
```
longquanzheng, line 125: nit: you may just create this map outside and share it with `persistence_loading_policy.LoadingTypePartialWithLock` branch.

integ/workflow/persistence_loading_policy/routers.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/indeedeng/iwf/service/common/ptr"
+	"log"
+	"net/http"
+)
+
+const (
+	WorkflowType                  = "persistence_loading_policy"
+	State1                        = "S1"
+	State2                        = "S2"
+	LoadingTypeAll                = "all"
```
longquanzheng, line 18: I think it's better to reuse the enum from idl. if you type 
```
iwfidl.ALL_WITHOUT_LOCKING
iwfidl.PARTIAL_WITHOUT_LOCKING
iwfidl.PARTIAL_WITH_EXCLUSIVE_LOCK
```

In golang, you can convert the enum from/to string directly:

```
  strVal = string(iwfidl.ALL_WITHOUT_LOCKING)

  persistenceLoadingType = iwfidl.PersistenceLoadingType(strVal)
```

longquanzheng (approved):
Mostly lgtm just some small comments on testing code pattern. very impressive job on first PR of Golang.

integ/util.go
longquanzheng: Actually nvm, they are default to true already.

integ/persistence_loading_policy_test.go
zklgame: Just tried it and found this way is really convenient!

integ/workflow/persistence_loading_policy/routers.go
zklgame: Great!

Issue #305, opened 2023-07-08
Support locking persistence for RPC 


longquanzheng:
Allow locking for persistence field (using validator in Temporal synchronous update api)

allow checking version in persistence field (same validator)

longquanzheng:
The first version will be supporting optimistic locking only

PR #320, opened 2023-08-07
Support None to not load any data for the persistence loading policy …

zklgame:
Support attributes loading None in RPC with code refactor

### DIFF
--- a/service/api/service.go
+++ b/service/api/service.go
@@ -460,9 +460,9 @@ func (s *serviceImpl) ApiV1WorkflowRpcPost(ctx context.Context, req iwfidl.Workf
 		}
 
 		// get loading keys for the search attributes
-		var loadingSAKeys []string
+		var saLoadingKeys []string
 		for _, requestedSA := range requestedSAs {
-			loadingSAKeys = append(loadingSAKeys, requestedSA.GetKey())
+			saLoadingKeys = append(saLoadingKeys, requestedSA.GetKey())
 		}
 
 		requestedSAs = append(requestedSAs, iwfidl.SearchAttributeKeyAndType{
@@ -481,7 +481,7 @@ func (s *serviceImpl) ApiV1WorkflowRpcPost(ctx context.Context, req iwfidl.Workf
 			allSearchAttributes = append(allSearchAttributes, sa)
 		}
 
-		searchAttributes = interpreter.LoadSearchAttributes(loadingSAKeys, allSearchAttributes)
+		searchAttributes = interpreter.LoadSearchAttributes(saLoadingKeys, allSearchAttributes)
 
 		daPolicy := req.GetDataAttributesLoadingPolicy()
 		if daPolicy.GetPersistenceLoadingType() == iwfidl.NONE {
--- a/service/api/service.go
+++ b/service/api/service.go
@@ -8,6 +8,7 @@ import (
 	"github.com/indeedeng/iwf/service/common/compatibility"
 	"github.com/indeedeng/iwf/service/common/urlautofix"
 	"github.com/indeedeng/iwf/service/common/utils"
+	"github.com/indeedeng/iwf/service/interpreter"
 	"io/ioutil"
 	"math"
 	"net/http"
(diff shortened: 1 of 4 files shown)


zklgame:
@longquanzheng The `s.client.QueryWorkflow` seems to be needed whatever the loading policies are since it gets the worker url.

longquanzheng:
> @longquanzheng The `s.client.QueryWorkflow` seems to be needed whatever the loading policies are since it gets other info like worker url, run id, timestamp.

ahh right. Maybe you can fix/improve this in the same PR(or separate if you prefer): https://github.com/indeedeng/iwf/blob/main/service/api/service.go#L112
We could always put this URL into memo/caching is not being requested for starting workflow. 
So that this RPC can try to use DescribeWF API to load workerURL and runId/timestamp when loading policy is none. (note changing the startWF code will only affect new workflow so you need to think about backward compatibility)

longquanzheng:
For testing, if we are able to optimize/improve it by avoid calling query workflow, I wonder if we can assert it in the test to make sure the query API is not being called. (we may need to think about some "hack" just for testing)

zklgame:
> > @longquanzheng The `s.client.QueryWorkflow` seems to be needed whatever the loading policies are since it gets other info like worker url, run id, timestamp.
> 
> ahh right. Maybe you can fix/improve this in the same PR(or separate if you prefer): https://github.com/indeedeng/iwf/blob/main/service/api/service.go#L112 We could always put this URL into memo/caching is not being requested for starting workflow. So that this RPC can try to use DescribeWF API to load workerURL and runId/timestamp when loading policy is none. (note changing the startWF code will only affect new workflow so you need to think about backward compatibility)
> 
> Because avoiding query is a very important optimization here. Query is very expensive. Otherwise none is useless for rpc.

Sounds like better making it in another PR since the new PR will include quite work:
1. get the workUrl, runId, timestamp from DescribeWorkflow for Temporal
2. support backward compatibility
3. test the calling of the query API

service/interpreter/persistence.go
```diff
@@ -127,17 +111,56 @@ func (am *PersistenceManager) LoadDataObjects(ctx UnifiedContext, loadingPolicy
...
-		return res.DataObjects
+		return LoadDataAttributes(partialLoadingKeys, am.GetAllDataObjects())
 	} else if loadingType == iwfidl.NONE {
 		return []iwfidl.KeyValue{}
 	} else {
 		panic("not supported loading type " + loadingType)
 	}
 }
 
+func LoadSearchAttributes(loadingKeys []string, allSearchAttributes []iwfidl.SearchAttribute) []iwfidl.SearchAttribute {
```
longquanzheng, line 122: As our/golang convention, the internal usage of a method/function should be kept as "private". In golang, the any var/method/func start with lower case is package private `loadSearchAttributes`. 
Also there is already another method `LoadSearchAttributes`, you may want to rename it to `internalLoadSearchAttributes` to be more clear.

service/interpreter/persistence.go
```diff
@@ -60,22 +60,16 @@ func RebuildPersistenceManager(provider WorkflowProvider,
...
-	for _, k := range request.Keys {
-		keyMap[k] = true
-	}
-	for key, value := range am.dataObjects {
-		if keyMap[key] || all {
-			res = append(res, value)
-		}
+		dataAttributes = am.GetAllDataObjects()
+	} else {
+		dataAttributes = LoadDataAttributes(request.Keys, am.GetAllDataObjects())
```
longquanzheng, line 68: I think it's a bit wastful to call `am.GetAllDataObjects()` before `LoadDataAttributes`. Because `GetAllDataObjects` will allocate a new slice/array to hold all the data attributes. Maybe `LoadDataAttributes` can just access to all the DAs directly

service/interpreter/persistence.go
longquanzheng: Or maybe `loadSearchAttributesByRequestedKeys`

service/api/service.go
```diff
@@ -432,56 +433,84 @@ func (s *serviceImpl) ApiV1WorkflowRpcPost(ctx context.Context, req iwfidl.Workf
...
 		var searchAttributes []iwfidl.SearchAttribute
 		var dataAttributes []iwfidl.KeyValue
 
-		requestedSAs := req.SearchAttributes
+		var requestedSAs []iwfidl.SearchAttributeKeyAndType
+
 		saPolicy := req.GetSearchAttributesLoadingPolicy()
-		if saPolicy.GetPersistenceLoadingType() != iwfidl.ALL_WITHOUT_LOCKING {
+		switch saPolicy.GetPersistenceLoadingType() {
+		case iwfidl.PARTIAL_WITH_EXCLUSIVE_LOCK:
```
longquanzheng, line 445: we need to return error for this case. this will be supported in https://github.com/indeedeng/iwf/issues/305

zklgame:
Close this one and to make a new one since it has been several weeks.

PR #337, opened 2023-08-25
Fix the loaded attributes with PRC memo

zklgame:
This MR did:
- correctly return the loaded attributes with RPC memo

### DIFF
--- a/service/api/service.go
+++ b/service/api/service.go
@@ -484,7 +484,7 @@ func (s *serviceImpl) tryPrepareRPCbyDescribe(ctx context.Context, req iwfidl.Wo
 	case iwfidl.ALL_WITHOUT_LOCKING:
 		requestedSAs = req.SearchAttributes
 	default:
-		return nil, s.handleError(fmt.Errorf("not supported search attributes loading policy: %s", saPolicy.GetPersistenceLoadingType()))
+		return nil, s.handleError(fmt.Errorf("not supported search attributes loading type: %s", saPolicy.GetPersistenceLoadingType()))
 	}
 
 	requestedSAs = append(requestedSAs, iwfidl.SearchAttributeKeyAndType{
@@ -535,7 +535,7 @@ func (s *serviceImpl) tryPrepareRPCbyDescribe(ctx context.Context, req iwfidl.Wo
 	case iwfidl.ALL_WITHOUT_LOCKING:
 		dataAttributes = allDataAttributes
 	default:
-		return nil, s.handleError(fmt.Errorf("not supported data attributes loading policy: %s", daPolicy.GetPersistenceLoadingType()))
+		return nil, s.handleError(fmt.Errorf("not supported data attributes loading type: %s", daPolicy.GetPersistenceLoadingType()))
 	}
 
 	attribute := response.SearchAttributes[service.SearchAttributeIwfWorkflowType]
--- a/service/api/service.go
+++ b/service/api/service.go
@@ -469,27 +469,23 @@ func (s *serviceImpl) tryPrepareRPCbyDescribe(ctx context.Context, req iwfidl.Wo
 	var requestedSAs []iwfidl.SearchAttributeKeyAndType
 	saPolicy := req.GetSearchAttributesLoadingPolicy()
 
-	if saPolicy.PersistenceLoadingType != nil {
-		switch saPolicy.GetPersistenceLoadingType() {
-		case iwfidl.PARTIAL_WITHOUT_LOCKING:
-			requestedSAKeys := map[string]bool{}
-			for _, saKey := range saPolicy.PartialLoadingKeys {
-				requestedSAKeys[saKey] = true
-			}
-			for _, sa := range req.SearchAttributes {
-				if requestedSAKeys[sa.GetKey()] {
-					requestedSAs = append(requestedSAs, sa)
-				}
-			}
-		case iwfidl.NONE:
-			requestedSAs = []iwfidl.SearchAttributeKeyAndType{}
-		case iwfidl.ALL_WITHOUT_LOCKING:
-			requestedSAs = req.SearchAttributes
-		default:
-			return nil, s.handleError(fmt.Errorf("not supported search attributes loading type: %s", saPolicy.GetPersistenceLoadingType()))
+	switch saPolicy.GetPersistenceLoadingType() {
+	case iwfidl.PARTIAL_WITHOUT_LOCKING:
+		requestedSAKeys := map[string]bool{}
+		for _, saKey := range saPolicy.PartialLoadingKeys {
+			requestedSAKeys[saKey] = true
 		}
-	} else {
+		for _, sa := range req.SearchAttributes {
+			if requestedSAKeys[sa.GetKey()] {
+				requestedSAs = append(requestedSAs, sa)
+			}
+		}
+	case iwfidl.NONE:
+		requestedSAs = []iwfidl.SearchAttributeKeyAndType{}
+	case iwfidl.ALL_WITHOUT_LOCKING, "":
 		requestedSAs = req.SearchAttributes
+	default:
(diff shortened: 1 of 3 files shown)


zklgame:
@longquanzheng ready to review now.

longquanzheng (changes requested):
this change looks good, but I think we also need to change line 406:
`if req.GetUseMemoForDataAttributes() {` to 
`if req.GetUseMemoForDataAttributes() || (use none loading policy)` otherwise it won't run into the code path

zklgame:
> this� change looks good, but I think we also need to change line 406: `if req.GetUseMemoForDataAttributes() {` to `if req.GetUseMemoForDataAttributes() || (use none loading policy)` otherwise it won't run into the code path

If adding the none checking here, it seems when the loading policy is NONE, some info will still be retrieved from memo (e.g., workerUrl). Is this the desired logic?

longquanzheng:
Yes, and we need to change the start api logic to always store worker url
into memo (even when memo is not enabled)

longquanzheng:
> > this� change looks good, but I think we also need to change line 406: `if req.GetUseMemoForDataAttributes() {` to `if req.GetUseMemoForDataAttributes() || (use none loading policy)` otherwise it won't run into the code path


also, I think the check should be even better:
`if (  useMemo || ( dataAttributePolicy == None && (searchAttributePolicy ==None ||  allSearchAttributes isEmpty )  ) `

Because in a lot of the case, user workflow don't have any search attributes. If they specify dataAttributeLodingPolicy is None then it's enough.

zklgame:
If to add  allSearchAttributes isEmpty in the if condition, then we have to get all the attributes from server in advance regardless the usage of the Memo. When there is actually using memo, it will incease unnecessary query to server to get the data attributes. It confuses me.

So I will not add this logic in the if condition.

longquanzheng:
> If to add allSearchAttributes isEmpty in the if condition, then we have to get all the attributes from server in advance regardless the usage of the Memo. When there is actually using memo, it will incease unnecessary query to server to get the data attributes. It confuses me.
> 
> So I will not add this logic in the if condition.

my bad -- I think I haven't make it clear and there are something tricky/bad in the API schema from me :/ 
this `searchAttributes` field in the RPCRequest https://github.com/indeedeng/iwf-idl/blob/c636d4b6ac1284d29a945aed9914baa37e6187da/iwf.yaml#L958 is providing all the defined search attributes of the workflow, from SDK (see example from Java SDK: https://github.com/indeedeng/iwf-java-sdk/blob/main/src/main/java/io/iworkflow/core/RpcInvocationHandler.java#L34 ) 

Ideally the field should be named `allWorkflowDefinedSearchAttributeTypes` to be clear, I made a wrong/unclear name there.

So with this field, we will be able to know if a workflow doens't have any search attribute in their schema. If it doesn't have any search attributes defined, then there is no value to use query to get search attribute. 


Side note for why `searchAttributes`/`allWorkflowDefinedSearchAttributeTypes` is necessary --- Temporal server/workflow doesn't have an easy way to store the user search attribute type. By default, an RPC need to return all the search attribute values with strongly typing. Therefore, the API protocol requires SDK to provide all the defined search attribute types in an RPC request.

zklgame:
Even if we could get all the search attributes from req, how about the data attributes? Currently if using memo, the data attribtues will be retrieved from memo too. But when the search attributes is empty, if we still run the getFromMemo logic, then the data attributes will be wrong. 

The logic here is getting more and more confusing to me as the discussion goes. So I think it would be much better and clear to see the real codes instead. What do you think?

longquanzheng:
Yeah that makes sense to me. We can keep it for next sync

zklgame:
I misread some message. Mofied the MR. But another test failed. I haven't found how to modify the tests.

service/api/service.go
```diff
@@ -79,13 +79,15 @@ func (s *serviceImpl) ApiV1WorkflowStartPost(ctx context.Context, req iwfidl.Wor
...
 		if startOptions.GetUseMemoForDataAttributes() {
-			workflowOptions.Memo = map[string]interface{}{
-				service.WorkerUrlMemoKey: iwfidl.EncodedObject{
-					Data: iwfidl.PtrString(req.IwfWorkerUrl), // this is hack to ensure all memos are with the same type
-				},
-			}
 			useMemo = true
 		}
+
+		// the workerUrl will always be stored in the memo regardless of `startOptions.GetUseMemoForDataAttributes()`
```
longquanzheng, line 85: I think we need to move it one more layer out. Even there is not startOptions, we still need to store the workerUrl into memo

service/api/service.go
```diff
@@ -454,19 +462,23 @@ func (s *serviceImpl) tryPrepareRPCbyDescribe(ctx context.Context, req iwfidl.Wo
 	var searchAttributes []iwfidl.SearchAttribute
 	var dataAttributes []iwfidl.KeyValue
 
-	requestedSAs := req.SearchAttributes
+	var requestedSAs []iwfidl.SearchAttributeKeyAndType
 	saPolicy := req.GetSearchAttributesLoadingPolicy()
-	if saPolicy.GetPersistenceLoadingType() != iwfidl.ALL_WITHOUT_LOCKING {
+	switch saPolicy.GetPersistenceLoadingType() {
+	case iwfidl.PARTIAL_WITH_EXCLUSIVE_LOCK, iwfidl.PARTIAL_WITHOUT_LOCKING:
```
longquanzheng, line 468: it shouldn't run into `PARTIAL_WITH_EXCLUSIVE_LOCK` because it will go to a different path `handleRpcBySynchronousUpdate` if needs locking. 
To be better understand, we should take this case out here -- you can return an error for this

service/api/service.go
```diff
@@ -454,19 +462,23 @@ func (s *serviceImpl) tryPrepareRPCbyDescribe(ctx context.Context, req iwfidl.Wo
...
 		requestedSAKeys := map[string]bool{}
 		for _, saKey := range saPolicy.PartialLoadingKeys {
 			requestedSAKeys[saKey] = true
 		}
-		requestedSAs = []iwfidl.SearchAttributeKeyAndType{}
```
longquanzheng, line 464: Removing this line is the reason that the test fail. 
In the original code, `requestedSAs` is reset here to empty, so that it will put the actual requestedSA into the list, based on the request.

service/api/service.go
```diff
@@ -454,19 +462,23 @@ func (s *serviceImpl) tryPrepareRPCbyDescribe(ctx context.Context, req iwfidl.Wo
...
 		for _, sa := range req.SearchAttributes {
 			if requestedSAKeys[sa.GetKey()] {
 				requestedSAs = append(requestedSAs, sa)
 			}
 		}
+	case iwfidl.NONE:
+		requestedSAs = []iwfidl.SearchAttributeKeyAndType{}
+	default:
```
longquanzheng, line 480: Nit: I think it's safer to explicitly handle the `ALL_WITHOUT_LOCKING` and return error for other cases that it doesn't understand

longquanzheng, line 480: Nit: I think it's safer to explicitly handle the ALL_WITHOUT_LOCKING and return error for other cases that it doesn't understand

zklgame: Modified it. But now we have to explicitly deal with the nil case.

longquanzheng (approved):

service/api/service.go
longquanzheng: alternatively, you can add a case PersistenceLoadingType("") to the `ALL_WITHOUT_LOCKING` branch:

```
case ALL_WITHOUT_LOCKING, PersistenceLoadingType("")
```

basically, `saPolicy.GetPersistenceLoadingType()` will return an empty string as `PersistenceLoadingType` when it's nil

zklgame: This sounds better!

PR #345, opened 2023-09-11
Support wait for state execution completion

duoertai:
not completed
TODO:
add state to wait for
add wait time in request
fix signal naming, etc
handle WaitForStateCompletion already running case

tested WaitForStateCompletion workflow and the api for happy case


service/api/cadence/client.go
```diff
@@ -113,6 +115,42 @@ func (t *cadenceClient) StartInterpreterWorkflow(ctx context.Context, options ap
 	return run.RunID, nil
 }
 
+func (t *cadenceClient) StartWaitForStateCompletionWorkflow(ctx context.Context, options api.StartWorkflowOptions, args ...interface{}) (runId string, err error) {
```
longquanzheng, line 118: This workflow should only need a workflowId as parameter. The ID reusepolicy can be just hardcoded to "DISALLOW_REUSE"

service/api/interfaces.go
```diff
@@ -31,6 +32,7 @@ type UnifiedClient interface {
 	Close()
 	errorHandler
 	StartInterpreterWorkflow(ctx context.Context, options StartWorkflowOptions, args ...interface{}) (runId string, err error)
+	StartWaitForStateCompletionWorkflow(ctx context.Context, options StartWorkflowOptions, args ...interface{}) (runId string, err error)
```
longquanzheng, line 35: We don't need this method here. The workflow should be started by `ApiV1WorkflowWaitForStateCompletion` instead

service/interpreter/interfaces.go
```diff
@@ -103,6 +104,12 @@ type WorkflowProvider interface {
 	GetBackendType() service.BackendType
 	GetLogger(ctx UnifiedContext) UnifiedLogger
 	NewInterpreterContinueAsNewError(ctx UnifiedContext, input service.InterpreterWorkflowInput) error
+	NewSelector(ctx UnifiedContext) UnifiedSelector
```
longquanzheng, line 107: we don't have to use selector. In this case, just adding `ReceiveBlocking` to `ReceiveChannel`  will work better(simpler)

service/interpreter/workflowImpl.go
```diff
@@ -788,3 +790,32 @@ func createUserWorkflowError(provider WorkflowProvider, message string) error {
 		message,
 	)
 }
+
+func WaitForStateCompletionWorkflowImpl(ctx UnifiedContext, provider WorkflowProvider, input service.WaitForStateCompletionWorkflowInput) (*service.WaitForStateCompletionWorkflowOutput, error) {
+	signalName := input.StateCompletionSignalId
```
longquanzheng, line 795: we don't need to pass this signalName. It can be just hardcoded in const.go

service/interpreter/temporal/workflowProvider.go
```diff
@@ -291,3 +298,39 @@ func (w *workflowProvider) GetUnhandledSignalNames(ctx interpreter.UnifiedContex
...
+	s.selector.Select(wfCtx)
+}
+
+func (s *selector) ReceiveSignalValueBlocking(ctx interpreter.UnifiedContext, signalName string) interface{} {
+	wfCtx, ok := ctx.GetContext().(workflow.Context)
+	if !ok {
+		panic("cannot convert to temporal workflow context")
+	}
+
+	var signalValue interface{}
```
longquanzheng, line 326: I am not entirely sure this will work if you pass an interface{} into the receive API. Maybe it will decoded into a string or something. But the better way (Go way) is to ask caller to provide a pointer like in `ReceiveAsync`

service/interpreter/workflowImpl.go
```diff
@@ -788,3 +790,32 @@ func createUserWorkflowError(provider WorkflowProvider, message string) error {
...
+	}
+
+	selector := provider.NewSelector(ctx)
+	signalValue := selector.ReceiveSignalValueBlocking(ctx, signalName)
+
+	if signalValue == nil {
+		return nil, createUserWorkflowError(provider, "state completion signal not found")
+	}
+
+	jsonData, err := json.Marshal(signalValue)
```
longquanzheng, line 807: Yeah like I mentioned above, the better way is to pass the pointer directly , like other APIs

service/interpreter/workflowImpl.go
```diff
@@ -788,3 +790,32 @@ func createUserWorkflowError(provider WorkflowProvider, message string) error {
...
+	if err != nil {
+		return nil, createUserWorkflowError(provider, "state completion signal value is not json")
+	}
+
+	var stateCompletionOutput iwfidl.StateCompletionOutput
+	err = json.Unmarshal(jsonData, &stateCompletionOutput)
+	if err != nil {
+		return nil, createUserWorkflowError(provider, "signal value is not StateCompletionOutput")
```
longquanzheng, line 815: If there is any error, it should be system error instead of user workflow error

service/api/cadence/client.go
longquanzheng: Also make sure to handle this error in service.go for already started error

service/api/service.go
```diff
@@ -117,6 +118,42 @@ func (s *serviceImpl) ApiV1WorkflowStartPost(ctx context.Context, req iwfidl.Wor
...
+		IwfWorkflowType:         "workflow type",
+		StateCompletionSignalId: "signal name",
+	})
+	if err != nil {
+		return nil, s.handleError(err)
+	}
+
+	s.logger.Info("Started workflow", tag.WorkflowID(req.WorkflowId), tag.WorkflowRunID(runId))
+
+	_, err = s.client.DescribeWorkflowExecution(ctx, workflowOptions.ID, runId, nil)
```
longquanzheng, line 139: this is not needed

service/api/service.go
```diff
@@ -117,6 +118,42 @@ func (s *serviceImpl) ApiV1WorkflowStartPost(ctx context.Context, req iwfidl.Wor
 	}, nil
 }
 
+func (s *serviceImpl) ApiV1WorkflowWaitForStateCompletion(ctx context.Context, req iwfidl.WorkflowWaitForStateCompletionRequest) (wresp *iwfidl.WorkflowWaitForStateCompletionResponse, retError *errors.ErrorAndStatus) {
+	defer func() { log.CapturePanic(recover(), s.logger, &retError) }()
+
+	workflowOptions := StartWorkflowOptions{
+		ID:        "__Iwf_system" + req.WorkflowId + "_" + req.StateExecutionId,
```
longquanzheng, line 125: Nit: `__Iwf_system` can be imoprted

service/interfaces.go
```diff
@@ -29,6 +29,11 @@ type (
 		ContinueAsNewInput ContinueAsNewInput `json:"continueAsNewInput"`
 	}
 
+	WaitForStateCompletionWorkflowInput struct {
```
longquanzheng, line 32: this is not needed

longquanzheng (changes requested):
Also need to add in activityImpl:

```
func StateApiExecute(...)
	defer func() {
		// if input.waiting { signalWithStart }
	}()
```

docker-compose/integ-dependencies.yml
```diff
@@ -120,7 +120,7 @@ services:
     depends_on:
       - cadence
     environment:
-      - CADENCE_CLI_ADDRESS=cadence:7933
+      - CADENCE_CLI_ADDRESS=cadence:7833
```
longquanzheng, line 123: changing this may be breaking the script which is for registering `default` domain

service/api/cadence/client.go
```diff
@@ -113,6 +114,28 @@ func (t *cadenceClient) StartInterpreterWorkflow(ctx context.Context, options ap
...
+	workflowOptions := client.StartWorkflowOptions{
+		ID:                           options.ID,
+		TaskList:                     options.TaskQueue,
+		WorkflowIDReusePolicy:        client.WorkflowIDReusePolicyRejectDuplicate,
+		ExecutionStartToCloseTimeout: 600 * time.Second, // TODO, make this configurable
+	}
+	run, err := t.cClient.StartWorkflow(ctx, workflowOptions, cadence.WaitforStateCompletionWorkflow)
+	if err != nil {
+		if t.IsWorkflowAlreadyStartedError(err) {
+			describeResponse, error := t.cClient.DescribeWorkflowExecution(ctx, workflowOptions.ID, "")
```
longquanzheng, line 127: the already started error contains the runId. Also within this file, there is no need to use `IsWorkflowAlreadyStartedError`

longquanzheng: actually we probably don't need to return runId?

service/interpreter/cadence/workflowProvider.go
```diff
@@ -2,20 +2,26 @@ package cadence
...
-	"time"
+	"go.uber.org/zap"
 )
 
 type workflowProvider struct {
 	threadCount        int
 	pendingThreadNames map[string]int
 }
 
+type selector struct {
```
longquanzheng, line 21: It's better to remove the unused code

service/interpreter/cadence/workflowProvider.go
```diff
@@ -254,3 +269,30 @@ func (w *workflowProvider) GetUnhandledSignalNames(ctx interpreter.UnifiedContex
...
+func (s *selector) Select(ctx interpreter.UnifiedContext) {
+	wfCtx, ok := ctx.GetContext().(workflow.Context)
+	if !ok {
+		panic("cannot convert to cadence workflow context")
+	}
+
+	s.selector.Select(wfCtx)
+}
+
+func (s *selector) ReceiveSignalValueBlocking(ctx interpreter.UnifiedContext, signalName string) interface{} {
```
longquanzheng, line 282: I believe we don't need to use selector. 

See how the ReceiveAsync is implemented:
```
func (t *cadenceReceiveChannel) ReceiveAsync(valuePtr interface{}) (ok bool) {
	return t.channel.ReceiveAsync(valuePtr)
}
```
It should be similar code

service/interpreter/temporal/workflowProvider.go
```diff
@@ -257,6 +264,15 @@ func (t *temporalReceiveChannel) ReceiveAsync(valuePtr interface{}) (ok bool) {
 	return t.channel.ReceiveAsync(valuePtr)
 }
 
+func (t *temporalReceiveChannel) ReceiveBlocking(ctx interpreter.UnifiedContext, valuePtr interface{}) (ok bool) {
```
longquanzheng, line 267: same here. The receiveBlocking should be simiar to receiveAsync

service/api/cadence/client.go
duoertai: is runId a required param in GetWorkflowResult?

service/interpreter/cadence/workflowProvider.go
```diff
@@ -254,3 +268,12 @@ func (w *workflowProvider) GetUnhandledSignalNames(ctx interpreter.UnifiedContex
 	}
 	return workflow.GetUnhandledSignalNames(wfCtx)
 }
+
+func (s *selector) Select(ctx interpreter.UnifiedContext) {
```
longquanzheng, line 272: Also need to remove this

service/interpreter/temporal/workflowProvider.go
```diff
@@ -2,21 +2,26 @@ package temporal
...
 	"go.temporal.io/sdk/workflow"
-	"time"
 )
 
 type workflowProvider struct {
 	threadCount        int
 	pendingThreadNames map[string]int
 }
 
+type selector struct {
```
longquanzheng, line 21: Same here -- need to remove

service/interpreter/temporal/workflowProvider.go
```diff
@@ -291,3 +305,11 @@ func (w *workflowProvider) GetUnhandledSignalNames(ctx interpreter.UnifiedContex
 	}
 	return workflow.GetUnhandledSignalNames(wfCtx)
 }
+
+func (s *selector) Select(ctx interpreter.UnifiedContext) {
```
longquanzheng, line 309: Same here

service/api/service.go
```diff
@@ -117,6 +120,39 @@ func (s *serviceImpl) ApiV1WorkflowStartPost(ctx context.Context, req iwfidl.Wor
...
+		TaskQueue:                s.taskQueue,
+		WorkflowExecutionTimeout: time.Duration(*req.WaitTimeSeconds) * time.Second,
+	}
+
+	runId, err := s.client.StartWaitForStateCompletionWorkflow(ctx, options)
+	if err != nil {
+		return nil, s.handleError(err)
+	}
+
+	waitTimeSeconds := 60 // default to 60 seconds
```
longquanzheng, line 138: I think we can just pass req.WaitTimeSeconds into TrimContextByTimeoutWithCappedDDL as a pointer

service/client/interfaces.go
```diff
@@ -0,0 +1,63 @@
+package client
```
longquanzheng, line 1: maybe even better to rename to `uclient`

service/interpreter/activityImpl.go
```diff
@@ -54,11 +57,28 @@ func StateApiWaitUntil(ctx context.Context, backendType service.BackendType, inp
...
-func StateApiExecute(ctx context.Context, backendType service.BackendType, input service.StateDecideActivityInput) (*iwfidl.WorkflowStateDecideResponse, error) {
+func StateApiExecute(ctx context.Context, backendType service.BackendType, input service.StateDecideActivityInput, shouldSendSignalOnCompletion bool) (*iwfidl.WorkflowStateDecideResponse, error) {
+	defer func() {
+		if shouldSendSignalOnCompletion {
+			unifiedCleint := env.GetUnifiedClient()
+			err := unifiedCleint.SignalWithStartWaitForStateCompletionWorkflow(
+				ctx, client.StartWorkflowOptions{
+					ID:                       service.IwfSystemConstPrefix + input.Request.Context.WorkflowId + "_" + *input.Request.Context.StateExecutionId,
+					TaskQueue:                env.GetTaskQueue(),
+					WorkflowExecutionTimeout: 600 * time.Second, // TODO: make it configurable
```
longquanzheng, line 72: This can be passed from the InterpreterWorkflow, they should use the same timeout

longquanzheng (approved):

PR #375, opened 2024-03-15
pin openapi-generator-cli to 6.6.0

duoertai:


duoertai:
@longquanzheng seems we use 6.6.0 before, and changing to 7.0.1 is a breaking change

longquanzheng (approved):

PR #376, opened 2024-03-19
Add start delay in workflow start options

duoertai:


duoertai:
Screenshot 2024-03-20 at 3 14 14 AM

tested with 10s delay

longquanzheng (approved):
maybe add a test?

duoertai:
> maybe add a test?

sry, merged before I saw this comment. I checked the current APIs, it seems that workflow start time is not included in any response(get, getWithWait, search), is there a good way to add test for this? Shall we add start time in the API response?

duoertai:
> maybe add a test?

let me try to use temporal describe api to get the actual start time

PR #380, opened 2024-05-10
Add detailed fields for error logs in api service

longquanzheng:


docker-compose/ci-cadence-dependencies.yml
```diff
@@ -23,24 +23,25 @@ services:
       - "9042:9042"
     networks:
       - testing-network
-  zookeeper:
```
longquanzheng, line 26: this is no longer available, see : https://github.com/uber/cadence/pull/5975

longquanzheng: cadence test is still failing, fixing the test env now

longquanzheng: Give up for now, still doesn't work: https://github.com/indeedeng/iwf/issues/381 for some reasons

lexi-indeed (approved):

lexihan0209 (approved):

PR #388, opened 2024-05-20
Use relative time(duration) for timer instead of absolute (firingUnixTimestampSeconds)

longquanzheng:


longquanzheng:
NOTE： This will need more work than this PR due to continueAsNew -- the duration value cannot be used in next run, because after continueAsNew, the workflow.now() has changed. We have to also write down the start time, or convert it to firingUnixTimestamp when the timer started, in order to pass to next run

PR #393, opened 2024-05-30
Use timer duration to replace firingTimestamp for timerCommand

longquanzheng:
Screenshot 2024-05-30 at 2 44 46 PM

### DIFF
--- a/service/interpreter/timerProcessor.go
+++ b/service/interpreter/timerProcessor.go
@@ -14,7 +14,9 @@ type TimerProcessor struct {
 	logger                          UnifiedLogger
 }
 
-func NewTimerProcessor(ctx UnifiedContext, provider WorkflowProvider, staleSkipTimerSignals []service.StaleSkipTimerSignal) *TimerProcessor {
+func NewTimerProcessor(
+	ctx UnifiedContext, provider WorkflowProvider, staleSkipTimerSignals []service.StaleSkipTimerSignal,
+) *TimerProcessor {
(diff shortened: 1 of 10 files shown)


service/interpreter/timerProcessor.go
```diff
@@ -144,3 +150,24 @@ func (t *TimerProcessor) AddTimers(stateExeId string, commands []iwfidl.TimerCom
...
+		if cmd.HasDurationSeconds() {
+			timerCommands = append(timerCommands, iwfidl.TimerCommand{
+				CommandId:                  cmd.CommandId,
+				FiringUnixTimestampSeconds: iwfidl.PtrInt64(now.Unix() + int64(cmd.GetDurationSeconds())),
+			})
+		} else {
+			timerCommands = append(timerCommands, iwfidl.TimerCommand{
+				CommandId:                  cmd.CommandId,
+				FiringUnixTimestampSeconds: cmd.FiringUnixTimestampSeconds,
+			})
```
mattschlitz, line 168: Maybe there's a technical reason for this but why don't we just append the original `cmd` here (like `timerCommands = append(timerCommands, cmd)`)?

mattschlitz (approved):

service/interpreter/timerProcessor.go
longquanzheng: Yeah I think that's simpler to do that. Thanks for pointing it out :D

PR #423, opened 2024-09-19
IWF-103: Require every commandId to exist for ANY_COMMAND_COMBINATION_COMPLETED

lwolczynski:


service/interpreter/activityImpl.go
```diff
@@ -224,12 +225,42 @@ func checkCommandRequestFromWaitUntilResponse(resp *iwfidl.WorkflowStateStartRes
 					return err
 				}
 			}
+			// Check if each command in the combinations has a matching command in one of the lists
+			if !areAllCommandCombinationsIdsValid(commandReq) {
+				return fmt.Errorf("ANY_COMMAND_COMBINATION_COMPLETED can only be used when every command has an commandId that is found in TimerCommands, SignalCommands or InterStateChannelCommands")
```
longquanzheng, line 230: `InterStateChannelCommands` is now renamed to `InternalChannelCommand`. There are still many places in server that hasn't updated. But all the API and error message that it returned should be using the new name for consistency.

longquanzheng:
The code looks good. 
It will be better to add a test. 
Unit tests are usually easier to write for those edge/error cases, but so far we only have integ test. 

In next meeting, probably we can do unit tests together. I have been lazy to write unit tests but I think it's important. 

But it's also possible to write integ tests -- 
* see how we test anyCommandCombination here: https://github.com/indeedeng/iwf/blob/main/integ/any_command_combination_test.go
* the workflow is defined in https://github.com/indeedeng/iwf/blob/main/integ/workflow/any_command_combination/routers.go#L27 
* in the workflow state code, it first return some invalid commands, and then on a retry, return the correct one: https://github.com/indeedeng/iwf/blob/main/integ/workflow/any_command_combination/routers.go#L61C2-L61C23

lwolczynski:
Added unit tests

longquanzheng (approved):

PR #424, opened 2024-09-23
IWF-70: allow headers to be set in Interpreter Config

### LINKED ISSUE #436: Add configuration option to allow headers to be sent on requests from iwf-server to app worker
When a client application worker receives requests from iwf-server, there should be an option in iwf to set custom headers to be sent in the request to the iwf worker.

samuel27m:
Closes #436


integ/util.go
```diff
@@ -67,6 +67,7 @@ type IwfServiceTestConfig struct {
 	MemoEncryption                   bool
 	DisableFailAtMemoIncompatibility bool // default to false so that we will fail at test
 	OptimizedVersioning              *bool
+	DefaultHeaders                   map[string]string
```
samuel27m, line 70: I saw that the previous implementation had integration tests that would check for the presence of the headers in every call, am I seeing that right? https://github.com/indeedeng/iwf/pull/356/files#diff-c3752f9aebfa9a670e62a5c0cc34c36bbfbfcbe849288ece0973806a34eac156R30-R33

I wanted to do something more flexible, where we would have a specific integration test for checking the headers passed in the configuration are being used, rather than having the route check it every time a call was made. 

I do need some of your guidance here, I can't figure out how to do the integration tests with this approach.

samuel27m:
@longquanzheng If you have some time tomorrow, let's chat tomorrow about this PR. I do need some guidance on the integration tests in the comment above.

integ/util.go
longquanzheng: The headers will be used when making calls for stateAPI. So for a specific integ tests, you will still need to build out a router. But it will be cleaner to have that specific router, is that what you meant?

samuel27m: That's what I'm thinking yes. Do you think that would be overkill for testing such a small feature?

longquanzheng: I think it’s a good idea. I have been lazy to do the right thing and the tests have become messy here

samuel27m: I have created the tests now, I struggled a bit navigating through the existing ones trying to use one as a template. I'm sure these can be better. We can have a chat tomorrow about them and you can point me where I've gone wrong! Thanks in advance! 🙌

integ/workflow/headers/routers.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/indeedeng/iwf/gen/iwfidl"
+	"github.com/indeedeng/iwf/service"
+	"log"
+	"net/http"
+)
+
+const (
+	WorkflowType = "basic"
+	State1       = "S1"
+	State2       = "S2"
```
longquanzheng, line 14: Nit: We can only have one state for this test

integ/workflow/headers/routers.go
```diff
@@ -0,0 +1,131 @@
...
+import (
+	"github.com/gin-gonic/gin"
+	"github.com/indeedeng/iwf/gen/iwfidl"
+	"github.com/indeedeng/iwf/service"
+	"log"
+	"net/http"
+)
+
+const (
+	WorkflowType = "basic"
```
longquanzheng, line 12: Nit: maybe "header"?

integ/workflow/headers/routers.go
```diff
@@ -0,0 +1,131 @@
...
+	if err := c.ShouldBindJSON(&req); err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		return
+	}
+	log.Println("received state start request, ", req)
+
+	context := req.GetContext()
+	if context.GetAttempt() <= 0 || context.GetFirstAttemptTimestamp() <= 0 {
+		panic("attempt and firstAttemptTimestamp should be greater than zero")
+	}
```
longquanzheng, line 44: you don't have to check this

integ/headers_test.go
```diff
@@ -0,0 +1,167 @@
...
+		DefaultHeaders: map[string]string{
+			headers.TestHeaderKey: headers.TestHeaderValue,
+		},
+	})
+	defer closeFunc2()
+
+	// start a workflow
+	sharedConfig := env.GetSharedConfig()
+	apiClient := iwfidl.NewAPIClient(&iwfidl.Configuration{
+		DefaultHeader: sharedConfig.Interpreter.InterpreterActivityConfig.DefaultHeader,
```
longquanzheng, line 66: I don't think we need this?

integ/headers_test.go
```diff
@@ -0,0 +1,167 @@
...
+	})
+	wfId := basic.WorkflowType + strconv.Itoa(int(time.Now().UnixNano()))
+	wfInput := &iwfidl.EncodedObject{
+		Encoding: iwfidl.PtrString("json"),
+		Data:     iwfidl.PtrString("test data"),
+	}
+	req := apiClient.DefaultApi.ApiV1WorkflowStartPost(context.Background())
+	startReq := iwfidl.WorkflowStartRequest{
+		WorkflowId:             wfId,
+		IwfWorkflowType:        basic.WorkflowType,
```
longquanzheng, line 81: We should start the header workflow here

integ/headers_test.go
```diff
@@ -0,0 +1,167 @@
...
+				MaximumAttempts:        iwfidl.PtrInt32(12),
+				MaximumIntervalSeconds: iwfidl.PtrInt32(12),
+			},
+			DecideApiRetryPolicy: &iwfidl.RetryPolicy{
+				InitialIntervalSeconds: iwfidl.PtrInt32(13),
+				BackoffCoefficient:     iwfidl.PtrFloat32(13),
+				MaximumAttempts:        iwfidl.PtrInt32(13),
+				MaximumIntervalSeconds: iwfidl.PtrInt32(13),
+			},
+		},
```
longquanzheng, line 114: we can skip all these

integ/util.go
longquanzheng: Sounds good. I went through it and left you some comments already. and we can go through them later

samuel27m:
Thanks for all the test feedback @longquanzheng! This should be good to review now, let me know if there's something that you think I could simplify further! Thanks in advance! 🙌

config/config.go
```diff
@@ -68,6 +68,7 @@ type (
 		// default is http://localhost:ApiConfig.Port
 		ApiServiceAddress                  string                              `json:"serviceAddress"`
 		DumpWorkflowInternalActivityConfig *DumpWorkflowInternalActivityConfig `json:"dumpWorkflowInternalActivityConfig"`
+		DefaultHeader                      map[string]string                   `json:"defaultHeader"`
```
longquanzheng, line 71: Not: headers (plural)

samuel27m: Good eye! Thanks, fixed now 🙌

integ/workflow/headers/routers.go
```diff
@@ -0,0 +1,92 @@
...
+	return &handler{
+		invokeHistory: make(map[string]int64),
+	}
+}
+
+// ApiV1WorkflowStartPost - for a workflow
+func (h *handler) ApiV1WorkflowStateStart(c *gin.Context) {
+	headerValue := c.GetHeader(TestHeaderKey)
+	if headerValue != TestHeaderValue {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "test header not found"})
```
longquanzheng, line 30: I think we need to return here after `c.JSON`. Otherwise the remaining code will override it.

integ/workflow/headers/routers.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+	}
+
+	c.JSON(http.StatusBadRequest, struct{}{})
+}
+
+func (h *handler) ApiV1WorkflowStateDecide(c *gin.Context) {
+	headerValue := c.GetHeader(TestHeaderKey)
+	if headerValue != TestHeaderValue {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "test header not found"})
```
longquanzheng, line 59: same here

integ/workflow/headers/routers.go
```diff
@@ -0,0 +1,92 @@
...
+	if err := c.ShouldBindJSON(&req); err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		return
+	}
+	log.Println("received state decide request, ", req)
+
+	if req.GetWorkflowType() == WorkflowType {
+		h.invokeHistory[req.GetWorkflowStateId()+"_decide"]++
+		if req.GetWorkflowStateId() == State1 {
+			// go to S2
```
longquanzheng, line 72: Nit: wrong comment

config/config.go
```diff
@@ -68,6 +68,7 @@ type (
 		// default is http://localhost:ApiConfig.Port
 		ApiServiceAddress                  string                              `json:"serviceAddress"`
 		DumpWorkflowInternalActivityConfig *DumpWorkflowInternalActivityConfig `json:"dumpWorkflowInternalActivityConfig"`
+		DefaultHeaders                     map[string]string                   `json:"defaultHeader"`
```
longquanzheng, line 71: ``json:"defaultHeaders"``

integ/workflow/headers/config.go
```diff
@@ -0,0 +1,4 @@
+package headers
+
+const TestHeaderKey = "integration-test-header"
+const TestHeaderValue = "integration-test-value"
```
longquanzheng, line 4: Nit: maybe constant.go? or just put it into `routers.go`

samuel27m:
Thanks for the comments! Should be all good now 👍

longquanzheng:
> Thanks for the comments! Should be all good now 👍

Did you forget to push the changes m?

samuel27m:
> renaming things to be more accurate

I did push them here: https://github.com/indeedeng/iwf/pull/424/commits/af66ce5b3d171e547c8ece47d8dbf89e20f130ca

Unless I forgot something ... 🤔

samuel27m:
Ok, now it should be good 😂

integ/workflow/headers/constants.go
```diff
@@ -0,0 +1 @@
+package headers
```
longquanzheng, line 1: Nite: we can remove it if it's empty

longquanzheng (approved):

PR #429, opened 2024-09-24
IWF-125: Limit error size return from API

lwolczynski:


service/interpreter/activityImpl.go
```diff
@@ -180,16 +180,29 @@ func composeHttpError(
...
-		errMsg = errMsg[:maxL] + "..."
 
 		maxL = len(responseBody)
 		if maxL > 50 {
 			maxL = 50
+			responseBody = responseBody[:maxL] + "..."
 		}
-		responseBody = responseBody[:maxL] + "..."
 		errType = "1st-attempt-failure"
+	} else {
```
longquanzheng, line 192: Nit: maybe nice to create a small func for the repeated code

service/interpreter/interfaces_mock.go
```diff
@@ -0,0 +1,701 @@
...
+	reflect "reflect"
+	time "time"
+
+	gomock "github.com/golang/mock/gomock"
+	iwfidl "github.com/indeedeng/iwf/gen/iwfidl"
+	service "github.com/indeedeng/iwf/service"
+)
+
+// MockActivityProvider is a mock of ActivityProvider interface.
+type MockActivityProvider struct {
```
longquanzheng, line 18: It’s better to name the file according so that it’s easier to find

lwolczynski: Is there a better way of creating this mock? I did it this way:
```
mockgen -source=service/interpreter/interfaces.go -package=interpreter -destination=service/interpreter/interfaces_mock.go 
```

longquanzheng: You can change the file name in the command or just rename the file in IDE. I am  not aware of other ways

lwolczynski: I'm confused why change the name at all. I mocked the `interface` file, so I thought `interfaces_mock` is a good name

lwolczynski: There are 13 mocked types in this file

longquanzheng: Oh you are right! My bad. I thought the file only contains the mock for activity provider which you used in the tests.

service/interpreter/activityImpl_test.go
```diff
@@ -88,3 +94,141 @@ func createCommands() ([]iwfidl.TimerCommand, []iwfidl.SignalCommand, []iwfidl.I
 	}
 	return validTimerCommands, validSignalCommands, internalCommands
 }
+
+func TestComposeHttpError_LocalActivity_LongErrorResponse(t *testing.T) {
```
stevo89519, line 98: Some of these test initializations are using duplicated logic as well. It'd be nice if these were consolidated. Maybe a function that takes the repeat length and returns the httpResp?

lwolczynski: I'll try to extract as much logic a to helper as I can

lwolczynski: Refactored

service/interpreter/activityImpl.go
lwolczynski: Refactored

stevo89519: Looks good

service/interpreter/activityImpl_test.go
stevo89519: Looks good

stevo89519 (approved):

longquanzheng (approved):

PR #431, opened 2024-09-25
Fix Cadence pipeline checks not starting

samuel27m:


docker-compose/ci-cadence-dependencies.yml
```diff
@@ -24,7 +24,7 @@ services:
     networks:
       - testing-network
   zookeeper:
-    image: wurstmeister/zookeeper:3.4.6
+    image: wurstmeister/zookeeper:latest
```
longquanzheng, line 27: Smart!!

samuel27m: Thanks! I do notice that the [`latest` version here was released 6 years ago](https://registry.hub.docker.com/layers/wurstmeister/zookeeper/latest/images/sha256-3f43f72cb2832e7a5fbed7f7dbcd0c43004357974d8f32555d101bd53e81e74f?context=explore)

Should we start using the [official zookeeper](https://registry.hub.docker.com/_/zookeeper) image instead? Is there a reason we're using this image instead?

longquanzheng: Yeah we should! 
Last time I spent a few minutes trying out but it didn’t work. I didn’t got time to. Investigate.

longquanzheng: Cadence has upgraded their docker compose to use the official one. I think we should do that and also upgrade the cadence image to the 1.x release

longquanzheng: Created https://github.com/indeedeng/iwf/issues/432

longquanzheng: The reason was that cadence was using that kafka image and I just followed it

PR #433, opened 2024-09-25
IWF-118: Refactor package structure

longquanzheng:


lwolczynski:
Do we need the "All tests" action? It seems to me like we are running ITs again Temporal and Cadence twice.

Once in "Temporal Integration Test" & "Candence Integration Test", and once in the "All tests"

longquanzheng:
> Do we need the "All tests" action? It seems to me like we are running ITs against Temporal and Cadence twice.
> 
> Once in "Temporal Integration Test" & "Candence Integration Test", and once in the "All tests"

Good question!
We need `all tests` because we need a single coverage report to calculate the coverage for all the tests that we have. 

So the temporal/cadence/unit tests are actually redudunt. They are needed because running this `all tests` is usually taking a ver long time. Sometimes it's much better to see the failure at unit tests/temporal/cadence tests

lwolczynski (approved):

PR #435, opened 2024-09-27
IWF-136: Support initial data attributes when starting workflow

lwolczynski:


integ/persistence_test.go
```diff
@@ -148,13 +155,31 @@ func doTestPersistenceWorkflow(
...
+				expectedDataAttribute,
+			},
 			WorkflowConfigOverride:   config,
 			UseMemoForDataAttributes: ptr.Any(useMemo),
 		},
 	}
 	_, httpResp, err := reqStart.WorkflowStartRequest(wfReq).Execute()
 	panicAtHttpError(err, httpResp)
 
+	// TODO: Fix the issue with running queryHandler before new workflow which Continues as New has started
```
longquanzheng, line 168: Looking into the code again, I now remember..
We intentionally set the query handler after the continueAsNew/dumpInternal activity. https://github.com/indeedeng/iwf/blob/main/service/interpreter/workflowImpl.go#L98

This is to ensure the correctness. If we set the query handler before that, the query handler could return empty data(since the loading hasn't completed), which will be incorrect response. 

So we would rather return server errors and let the client retry later.

longquanzheng: I think we should add a comment at https://github.com/indeedeng/iwf/blob/main/service/interpreter/workflowImpl.go#L98. to remind people why this is not before https://github.com/indeedeng/iwf/blob/main/service/interpreter/workflowImpl.go#L63

longquanzheng (approved):

integ/persistence_test.go
longquanzheng: For this test, we can add a comment to explain why a sleep is necessary

integ/persistence_test.go
```diff
@@ -148,13 +155,31 @@ func doTestPersistenceWorkflow(
...
+	time.Sleep(time.Millisecond * 10)
```
longquanzheng, line 169: Looks like the test failed. I looked at my other sleeps for the same reasons, I used 1 second for safety. Also it's better to only sleep for continueAsNew cases. (the test will run faster that way)

integ/persistence_test.go
```diff
@@ -148,13 +155,34 @@ func doTestPersistenceWorkflow(
...
 		},
 	}
 	_, httpResp, err := reqStart.WorkflowStartRequest(wfReq).Execute()
 	panicAtHttpError(err, httpResp)
 
+	// Config is only present for continueAsNew tests
+	if config != nil {
+		// TODO: Fix the issue with running queryHandler before new workflow which Continues as New has started
+		time.Sleep(time.Millisecond * 1000)
+	}
```
lwolczynski, line 172: Bumped the sleep to 1 second and wrapped it in a check if config is present

lwolczynski: @longquanzheng

integ/persistence_test.go
```diff
@@ -148,13 +156,39 @@ func doTestPersistenceWorkflow(
...
+		for {
+			if err == nil || retryCount >= 5 {
+				break
+			}
+			// Loading data to a continuedAsNew workflow might take a few seconds thus retry mechanism is needed
+			time.Sleep(time.Millisecond * 1000)
+			retryCount += 1
+			queryResult, httpResp, err = getDataAttributes(initReqQry, wfId, expectedDataAttribute, useMemo)
+		}
+	}
```
lwolczynski, line 186: @longquanzheng what are your thoughts about this?

longquanzheng: That's fine. I was thinking about the same way to workaround. 

However, ideally we can do this in api service for everyone (that will saves all the sleeping in the tests).

But it's better to do in a separate PR (need to figure out checking error types, and backoff strategy and refactoring all the tests to remove the sleep).

Could you also raise another ticket for this?

lwolczynski: Will do!

PR #440, opened 2024-10-04
Draft: IWF-138: Add stateId to WorkflowWaitForStateCompletion

lwolczynski:
Tests pending

### DIFF
--- a/service/api/service.go
+++ b/service/api/service.go
@@ -158,7 +158,7 @@ func (s *serviceImpl) ApiV1WorkflowWaitForStateCompletion(
 
 	var workflowId string
 	if req.WaitForKey != nil {
-		workflowId = service.IwfSystemConstPrefix + req.WorkflowId + "_" + *req.WaitForKey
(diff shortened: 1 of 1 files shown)


longquanzheng:
We also need to change the workflow code to signalWithStart the right one

longquanzheng (approved):

integ/wait_for_state_completion_test.go
```diff
@@ -74,49 +74,75 @@ func doTestWaitForStateCompletion(
...
+		var output iwfidl.WorkflowWaitForStateCompletionResponse
+		defer httpResp.Body.Close()
+		err = json.NewDecoder(httpResp.Body).Decode(&output)
+		if err != nil {
+			log.Fatalf("Failed to decode the response: %v", err)
+		}
 	}
-	assertions.Equal("S1-1", output.StateCompletionOutput.CompletedStateExecutionId)
+
+	assertions2 := assert.New(t)
```
longquanzheng, line 129: Nit: what happened to assertion 1? I believe we can reuse it

lwolczynski: You're right. Fixed

PR #442, opened 2024-10-08
IWF-186: Upgrade Temporal SDK

lwolczynski:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #issue_number


integ/any_command_combination_test.go
```diff
@@ -108,7 +108,7 @@ func doTestAnyCommandCombinationWorkflow(t *testing.T, backendType service.Backe
 	panicAtHttpError(err, httpResp)
 
 	// skip the timer for S1
-	time.Sleep(time.Second * 2) // wait for a second so that timer is ready to be skipped
+	time.Sleep(time.Second * 5) // wait for a second so that timer is ready to be skipped
```
lwolczynski, line 111: I had to bump this sleep time from 2 to 5 seconds. Otherwise, the test was flaky and sometimes panicked at line 118 due to a 400 error received

lwolczynski:
To review: `TestAnyCommandCombinationWorkflowTemporal` test passes locally, fails in Gitlab CI

integ/locking_test.go
```diff
@@ -116,7 +116,7 @@ func doTestLockingWorkflow(t *testing.T, backendType service.BackendType, config
 					ValueType: iwfidl.INT.Ptr(),
 				},
 			}
-			time.Sleep(time.Second * 1)
+			time.Sleep(time.Second * 2)
```
lwolczynski, line 119: Test passes locally with no change. I had to bump it to make the test pass in Github CI

integ/any_command_combination_test.go
```diff
@@ -108,7 +108,7 @@ func doTestAnyCommandCombinationWorkflow(t *testing.T, backendType service.Backe
 	panicAtHttpError(err, httpResp)
 
 	// skip the timer for S1
-	time.Sleep(time.Second * 2) // wait for a second so that timer is ready to be skipped
+	time.Sleep(time.Second * 5) // wait for a few seconds so that timer is ready to be skipped
```
lwolczynski, line 111: Same here

service/interpreter/stateRequestQueue.go
```diff
@@ -58,6 +58,17 @@ func (srq *StateRequestQueue) GetAllStateStartRequests() []iwfidl.StateMovement
 	return res
 }
 
+func (srq *StateRequestQueue) GetAllStateResumeStates() []service.StateExecutionResumeInfo {
```
longquanzheng, line 61: Nit: GetAllStateResumeRequests()

integ/locking_test.go
longquanzheng: I am still not sure why we need more sleep. I remember there is a new behavior in temporal cloud that they now enforce a backoff of 1s for each continueAsNew to prevent having too fast continueAsNew which cause hot partition in server. Maybe that's the cause. we can dig into the release logs to confirm, or maybe ask them

longquanzheng: actually the minimum backoff has been there for a long time: https://github.com/temporalio/temporal/blame/fad6bdc0e9c0911f28829f3c47285357554e2567/service/history/workflow/mutable_state_impl.go#L1703

lwolczynski: I'm looking into this, but it's not easy to troubleshoot since it passes locally

longquanzheng: I just looked into that, looks like there is a test panic at timeout for the cadence cases

integ/do_nothing_test.go
```diff
@@ -0,0 +1,14 @@
+package integ
```
longquanzheng, line 1: nit: `do_nothing_test` -> `run_server_for_debug_test`

service/api/temporal/reset.go
```diff
@@ -117,7 +117,7 @@ OuterLoop:
 		}
 		for _, e := range resp.GetHistory().GetEvents() {
 			if e.GetEventType() == enums.EVENT_TYPE_WORKFLOW_TASK_COMPLETED {
-				if e.GetEventTime().UnixNano() >= earliestTime {
+				if e.GetEventTime().GetSeconds()*1000*1000*1000+int64(e.GetEventTime().GetNanos()) >= earliestTime {
```
longquanzheng, line 120: nit: maybe create a new util method somewhere:
```
util.ToNanoSeconds(e.GetEventTime())

```

service/api/temporal/client.go
```diff
@@ -273,7 +273,7 @@ func (t *temporalClient) DescribeWorkflowExecution(
 		Status:                   status,
 		SearchAttributes:         searchAttributes,
 		Memos:                    memo,
-		WorkflowStartedTimestamp: resp.GetWorkflowExecutionInfo().GetStartTime().Unix(),
+		WorkflowStartedTimestamp: resp.GetWorkflowExecutionInfo().GetStartTime().GetSeconds()*1000*1000*1000 + int64(resp.GetWorkflowExecutionInfo().GetStartTime().GetNanos()),
```
longquanzheng, line 276: same here for the util method

longquanzheng (approved):

Makefile
```diff
@@ -183,10 +183,12 @@ ci-temporal-integ-test:
 	$Q go test -v ./integ  -cover -coverprofile coverage.out -coverpkg ./service/... -search=false -cadence=false -dependencyWaitSeconds=60
 
 ci-all-tests:
-	$Q go test -v ./... -cover -coverprofile coverage.out -coverpkg ./service/...
+ 	# Fails CI when used with -coverprofile flag due to tests that panic; see https://go.dev/doc/build-cover#panicprof
+ 	# $Q go test -v ./... -timeout 15m -cover -coverprofile coverage.out -coverpkg ./service/...
+	$Q go test -v ./... -timeout 15m
```
lwolczynski, line 188: @longquanzheng CI issue is caused by Go upgrade from 1.19 to 1.22

> Beginning in Go 1.20, Go supports collection of coverage profiles from applications and from integration tests, larger and more complex tests for Go programs.



> If my program panics, will coverage data be written?[¶](https://go.dev/doc/build-cover#panicprof)
> Programs built with go build -cover will only write out complete profile data at the end of execution if the program invokes os.Exit() or returns normally from main.main. If a program terminates in an unrecovered panic, or if the program hits a fatal exception (such as a segmentation violation, divide by zero, etc), profile data from statements executed during the run will be lost.

When tests panic, the task returns "Error 1" which fails the Github Action. We should fix it by refactoring tests, but I removed the coverage options to unblock now for

https://go.dev/doc/build-cover#panicprof

longquanzheng: https://github.com/indeedeng/iwf/issues/446

PR #447, opened 2024-10-11
IWF-132: Use firstRunId instead workflowId when naming WaitForStateExecutionCompletion workflow

### LINKED ISSUE #404: Workaround for WaitForStateExecutionCompletion limitation -- use firstRunId instead workflowId for waitFor workflow

lwolczynski:
### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related Issue
Closes #404

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -31,6 +31,15 @@ type (
 		// omitRpcInputOutputInHistory is the flag to omit rpc input/output in history
 		// the input/output is only for debugging purpose but could be too expensive to store
 		OmitRpcInputOutputInHistory *bool `yaml:"omitRpcInputOutputInHistory"`
+		// WaitForStateCompletionMigration is used to control naming of the continuedAsNew workflows
+		WaitForStateCompletionMigration WaitForStateCompletionMigration `yaml:"waitForStateCompletionMigration"`
+	}
+
+	WaitForStateCompletionMigration struct {
+		// expected values: old/both/new; defaults to 'old'
+		SignalWithStartOn string `yaml:"signalWithStartOn"`
+		// expected values: old/new; defaults to 'old'
+		WaitForOn string `yaml:"waitForOn"`
 	}
 
 	Interpreter struct {
--- a/config/config.go
+++ b/config/config.go
@@ -31,7 +31,7 @@ type (
 		// omitRpcInputOutputInHistory is the flag to omit rpc input/output in history
 		// the input/output is only for debugging purpose but could be too expensive to store
 		OmitRpcInputOutputInHistory *bool `yaml:"omitRpcInputOutputInHistory"`
-		// WaitForStateCompletionMigration is used to control naming of the continuedAsNew workflows
+		// WaitForStateCompletionMigration is used to control naming of the not continuedAsNew workflows
 		WaitForStateCompletionMigration WaitForStateCompletionMigration `yaml:"waitForStateCompletionMigration"`
 	}
 
--- a/config/config.go
+++ b/config/config.go
@@ -136,3 +136,17 @@ func GetApiServiceAddressWithDefault(config Config) string {
 	}
 	return fmt.Sprintf("http://localhost:%v", config.Api.Port)
 }
+
+func GetSignalWithStartOnWithDefault(config Config) string {
+	if config.Api.WaitForStateCompletionMigration.SignalWithStartOn != "" {
+		return config.Api.WaitForStateCompletionMigration.SignalWithStartOn
+	}
+	return "old"
+}
+
+func GetWaitForOnWithDefault(config Config) string {
+	if config.Api.WaitForStateCompletionMigration.WaitForOn != "" {
+		return config.Api.WaitForStateCompletionMigration.WaitForOn
+	}
+	return "old"
+}
--- a/config/config.go
+++ b/config/config.go
@@ -130,23 +130,23 @@ func NewConfig(configPath string) (*Config, error) {
 	return config, nil
 }
 
-func GetApiServiceAddressWithDefault(config Config) string {
-	if config.Interpreter.InterpreterActivityConfig.ApiServiceAddress != "" {
-		return config.Interpreter.InterpreterActivityConfig.ApiServiceAddress
+func (c Config) GetApiServiceAddressWithDefault() string {
+	if c.Interpreter.InterpreterActivityConfig.ApiServiceAddress != "" {
+		return c.Interpreter.InterpreterActivityConfig.ApiServiceAddress
 	}
-	return fmt.Sprintf("http://localhost:%v", config.Api.Port)
+	return fmt.Sprintf("http://localhost:%v", c.Api.Port)
 }
 
-func GetSignalWithStartOnWithDefault(config Config) string {
-	if config.Api.WaitForStateCompletionMigration.SignalWithStartOn != "" {
-		return config.Api.WaitForStateCompletionMigration.SignalWithStartOn
+func (c Config) GetSignalWithStartOnWithDefault() string {
+	if c.Api.WaitForStateCompletionMigration.SignalWithStartOn != "" {
+		return c.Api.WaitForStateCompletionMigration.SignalWithStartOn
 	}
 	return "old"
 }
 
-func GetWaitForOnWithDefault(config Config) string {
-	if config.Api.WaitForStateCompletionMigration.WaitForOn != "" {
-		return config.Api.WaitForStateCompletionMigration.WaitForOn
+func (c Config) GetWaitForOnWithDefault() string {
+	if c.Api.WaitForStateCompletionMigration.WaitForOn != "" {
+		return c.Api.WaitForStateCompletionMigration.WaitForOn
 	}
 	return "old"
 }
--- a/service/api/service.go
+++ b/service/api/service.go
@@ -162,7 +162,7 @@ func (s *serviceImpl) ApiV1WorkflowWaitForStateCompletion(
 	waitForOn := env.GetSharedConfig().Api.WaitForStateCompletionMigration.WaitForOn
 
 	if waitForOn == "old" {
-		workflowId = utils.GetWorkflowIdForWaitForStateExecution(req.WorkflowId, *req.StateExecutionId, *req.WaitForKey, *req.StateId)
+		workflowId = utils.GetWorkflowIdForWaitForStateExecution(req.WorkflowId, req.StateExecutionId, req.WaitForKey, req.StateId)
 	} else { // waitForOn == "new"
 		var parentId string
 		if s.client.GetBackendType() == service.BackendTypeTemporal { // Temporal
@@ -175,7 +175,7 @@ func (s *serviceImpl) ApiV1WorkflowWaitForStateCompletion(
 			parentId = req.WorkflowId
 		}
 
-		workflowId = utils.GetWorkflowIdForWaitForStateExecution(parentId, *req.StateExecutionId, *req.WaitForKey, *req.StateId)
+		workflowId = utils.GetWorkflowIdForWaitForStateExecution(parentId, req.StateExecutionId, req.WaitForKey, req.StateId)
 	}
 
 	options := uclient.StartWorkflowOptions{
--- a/service/api/service.go
+++ b/service/api/service.go
@@ -158,15 +158,15 @@ func (s *serviceImpl) ApiV1WorkflowWaitForStateCompletion(
 
 	var parentId, currentWorkflowId, _ string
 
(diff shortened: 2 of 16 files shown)


service/api/service.go
```diff
@@ -156,15 +156,30 @@ func (s *serviceImpl) ApiV1WorkflowWaitForStateCompletion(
...
-	var workflowId string
+	var parentWfId, currentWorkflowId, _ string
+
+	response, err := s.client.DescribeWorkflowExecution(ctx, req.GetWorkflowId(), "", nil)
+	if err != nil {
+		return nil, s.handleError(err, WorkflowWaitForStateCompletionApiPath, req.WorkflowId)
+	}
+
+	if response.FirstRunId == "" {
+		parentWfId = req.WorkflowId // Cadence
```
longquanzheng, line 167: If we know cadence doesn't have this, we can avoid calling the `DescribeWorkflowExecution` API for cadence. 

Although this API is really fast (it's backed by caching on cadence/temporal) but still nice to do so if this is not too much. (we do have cadence users in OSS, some ppl just perfer using cadence)

service/api/service.go
```diff
@@ -156,15 +156,30 @@ func (s *serviceImpl) ApiV1WorkflowWaitForStateCompletion(
...
+		_ = service.IwfSystemConstPrefix + parentWfId + "_" + *req.StateId + "_" + *req.WaitForKey
 	} else {
-		workflowId = service.IwfSystemConstPrefix + req.WorkflowId + "_" + *req.StateExecutionId
+		currentWorkflowId = service.IwfSystemConstPrefix + parentWfId + "_" + *req.StateExecutionId
+		_ = service.IwfSystemConstPrefix + parentWfId + "_" + *req.StateExecutionId
 	}
 
 	options := uclient.StartWorkflowOptions{
-		ID:        workflowId,
+		// TODO: Switch currentWorkflowId to _ (after renaming) after a short amount of time to ensure backward compatibility
```
longquanzheng, line 181: Ohh interesting! I was thinking of making two deployment -- first deploy the workflowImpl code (which signalWIthStart both), after some time, deploy the api service. 
But what you put here made me feel like it's nice to use a config? 
Potentially using the `Api ApiConfig `yaml:"api"`` in https://github.com/indeedeng/iwf/blob/main/config/config.go 
```
waitForStateCompletionMigration:
   SignalWithStartOn: both/new/old
   WaitForOn:old/new
```

service/api/service.go
```diff
@@ -156,15 +156,30 @@ func (s *serviceImpl) ApiV1WorkflowWaitForStateCompletion(
...
+
+	response, err := s.client.DescribeWorkflowExecution(ctx, req.GetWorkflowId(), "", nil)
+	if err != nil {
+		return nil, s.handleError(err, WorkflowWaitForStateCompletionApiPath, req.WorkflowId)
+	}
+
+	if response.FirstRunId == "" {
+		parentWfId = req.WorkflowId // Cadence
+	} else {
+		parentWfId = response.FirstRunId // Temporal
```
longquanzheng, line 169: very nit: parentId? it's not workflowId Lol.

service/interpreter/workflowImpl.go
```diff
@@ -763,18 +763,31 @@ func invokeStateExecute(
 		// this is not a problem because the signalWithStart will be very fast and highly available
 		unifiedClient := env.GetUnifiedClient()
 
-		var workflowId string
+		var parentWfId, currentWorkflowId, newWorkflowId string
+
+		response, err := unifiedClient.DescribeWorkflowExecution(context.Background(), executionContext.WorkflowId, "", nil)
```
longquanzheng, line 768: we should get the firstRunId from workflowContext.
Basically, temporal workflow provides the runtime context via context. (similarly iWF workflowState has Context). 
The runtime context has access to this runId directly without making an API call.

service/interpreter/workflowImpl.go
```diff
@@ -784,6 +797,23 @@ func invokeStateExecute(
 			// panic will let the workflow task will retry until the signal is sent
 			panic(fmt.Errorf("failed to signal on completion %w", err))
 		}
+
+		if response.FirstRunId != "" { // Temporal
+			// Start WaitForStateCompletionWorkflow with a new name to ensure smooth transition
+			err = unifiedClient.SignalWithStartWaitForStateCompletionWorkflow(
```
longquanzheng, line 803: Nit: we can refactor the code to a common method

service/api/service.go
longquanzheng: In workflowImpl, we have access to ApiConfig via env like this: https://github.com/indeedeng/iwf/blob/main/service/interpreter/continueAsNewer.go#L57

service/api/service.go
lwolczynski: 👍 done

service/api/service.go
lwolczynski: That's a good idea. I'll implement it

service/interpreter/workflowImpl.go
lwolczynski: I'm not sure if this is what you meant by workflowContext. I added FirstRunID to WorkflowInfo interface and use it this way:
```
provider.GetWorkflowInfo(ctx).FirstRunID
```

I think that's what you meant

service/api/service.go
lwolczynski: Added GetBackendType method to UnifiedClient

service/interpreter/workflowImpl.go
lwolczynski: Done

service/api/service.go
lwolczynski: Done

service/interpreter/workflowImpl.go
```diff
@@ -808,6 +806,38 @@ func invokeStateExecute(
...
+	}
+	return nil, service.FailureStateExecutionStatus, convertStateApiActivityError(provider, err)
+}
+
+func signalWithStartWorkflowId(state iwfidl.StateMovement, parentId string, executionContext iwfidl.Context) string {
+	if state.WaitForKey != nil {
+		return service.IwfSystemConstPrefix + parentId + "_" + state.StateId + "_" + *state.WaitForKey
+	} else {
+		return service.IwfSystemConstPrefix + parentId + "_" + *executionContext.StateExecutionId
+	}
```
longquanzheng, line 821: Nit: it would be nice to move this to a common place to share with api service: `GetWorkflowIdForWaitForStateExecution(parentId, stateExeId, waitForKey, stateId)`

service/interpreter/workflowImpl.go
```diff
@@ -808,6 +806,38 @@ func invokeStateExecute(
...
+			TaskQueue:                env.GetTaskQueue(),
+			WorkflowExecutionTimeout: 60 * time.Second, // timeout doesn't matter here as it will complete immediate with the signal
+		},
+		iwfidl.StateCompletionOutput{})
+	if err != nil && !unifiedClient.IsWorkflowAlreadyStartedError(err) {
+		// WorkflowAlreadyStartedError is returned when the started workflow is closed and the signal is not sent
+		// panic will let the workflow task will retry until the signal is sent
+		panic(fmt.Errorf("failed to signal on completion %w", err))
+	}
+	return err
```
longquanzheng, line 838: I think this func doesn't need to return anything(error).

It's a bit tricky here that we need to expand on later about "non determinism and how workflow task is executed". 

For here, we intentionally swallow the  error on "alreadyStarted", and panic on any other errors. By panicing, it will just retry automatically. 

We don't want to proceed to other states for any other errors.

service/api/service.go
```diff
@@ -199,6 +215,14 @@ func (s *serviceImpl) ApiV1WorkflowWaitForStateCompletion(
 	}, nil
 }
 
+func WorkflowWaitForStateId(req iwfidl.WorkflowWaitForStateCompletionRequest, parentId string) string {
```
longquanzheng, line 218: 👍

service/interpreter/workflowImpl.go
lwolczynski: Done

service/interpreter/workflowImpl.go
lwolczynski: Fixed the issue

service/interpreter/workflowImpl.go
```diff
@@ -763,28 +764,35 @@ func invokeStateExecute(
...
+
+			// Start WaitForStateCompletionWorkflow with a new name to ensure smooth transition
+			err = signalWithStart(unifiedClient, workflowId)
+			if err != nil && !unifiedClient.IsWorkflowAlreadyStartedError(err) {
+				// WorkflowAlreadyStartedError is returned when the started workflow is closed and the signal is not sent
+				// panic will let the workflow task will retry until the signal is sent
+				panic(fmt.Errorf("failed to signal on completion %w", err))
+			}
 		}
 	}
```
lwolczynski, line 794: Refactored this from two nested if statements into two separate ifs with more complicated expressions

service/api/service.go
```diff
@@ -157,10 +158,24 @@ func (s *serviceImpl) ApiV1WorkflowWaitForStateCompletion(
...
+			response, err := s.client.DescribeWorkflowExecution(ctx, req.GetWorkflowId(), "", nil)
+			if err != nil {
+				return nil, s.handleError(err, WorkflowWaitForStateCompletionApiPath, req.WorkflowId)
+			}
+			parentId = response.FirstRunId
+		} else { // Cadence
+			parentId = req.WorkflowId
+		}
+
+		workflowId = utils.GetWorkflowIdForWaitForStateExecution(parentId, *req.StateExecutionId, *req.WaitForKey, *req.StateId)
```
longquanzheng, line 178: `*req.WaitForKey` will assume that field is always present (otherwise it will be a panic -- access to nil value). I think only StateExecutionId is required. So we probably need to pass pointer into the method and let the method do a nil check

lwolczynski: Fixed

service/interpreter/workflowImpl.go
```diff
@@ -763,28 +765,35 @@ func invokeStateExecute(
...
+		// signalWithStart with legacy workflowId (containing parent workflowId)
+		if provider.GetBackendType() == service.BackendTypeCadence ||
+			(provider.GetBackendType() == service.BackendTypeTemporal && (signalWithStartOn == "old" || signalWithStartOn == "both")) {
+			workflowId := utils.GetWorkflowIdForWaitForStateExecution(executionContext.WorkflowId, executionContext.StateExecutionId, state.WaitForKey, &state.StateId)
+
+			err = signalWithStart(unifiedClient, workflowId)
+			if err != nil && !unifiedClient.IsWorkflowAlreadyStartedError(err) {
+				// WorkflowAlreadyStartedError is returned when the started workflow is closed and the signal is not sent
+				// panic will let the workflow task will retry until the signal is sent
+				panic(fmt.Errorf("failed to signal on completion %w", err))
```
longquanzheng, line 779: Nit: it will look slightly better to move this error handling into the signalWithStart method

longquanzheng (approved):

config/config.go
```diff
@@ -127,3 +136,17 @@ func GetApiServiceAddressWithDefault(config Config) string {
...
 }
+
+func GetSignalWithStartOnWithDefault(config Config) string {
+	if config.Api.WaitForStateCompletionMigration.SignalWithStartOn != "" {
+		return config.Api.WaitForStateCompletionMigration.SignalWithStartOn
+	}
+	return "old"
+}
+
+func GetWaitForOnWithDefault(config Config) string {
```
longquanzheng, line 147: Nit: I actually should have done it this using struct member method instead:
```
func (c Config) GetSignalWithStartOnWithDefault() string{
    // the same code here
}
```
And to use it:
```
config.GetSignalWithStartOnWithDefault()
```

This is very common pattern in Golang, I don't know why I didn't do that way 😂

lwolczynski: Done

service/interpreter/workflowImpl.go
lwolczynski: Addressed

config/config.go
```diff
@@ -31,6 +31,15 @@ type (
 		// omitRpcInputOutputInHistory is the flag to omit rpc input/output in history
 		// the input/output is only for debugging purpose but could be too expensive to store
 		OmitRpcInputOutputInHistory *bool `yaml:"omitRpcInputOutputInHistory"`
+		// WaitForStateCompletionMigration is used to control naming of the continuedAsNew workflows
```
longquanzheng, line 34: typo: not `continuedAsNew`

PR #448, opened 2024-10-14
IWF-137: Allow to use separate persistency loading policy for waitUntil

### LINKED ISSUE #387: Allow to use separate persistency loading policy for waitUntil
currently it's using the same for both(waitUntil/execute)

samuel27m:
### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related Issue
Closes #387


service/common/compatibility/stateOptions.go
```diff
@@ -45,14 +45,60 @@ func GetDecideApiRetryPolicy(stateOptions *iwfidl.WorkflowStateOptions) *iwfidl.
...
+	if stateOptions.HasDataObjectsLoadingPolicy() {
+		return stateOptions.DataObjectsLoadingPolicy
+	}
+
+	if stateOptions.HasDataAttributesLoadingPolicy() {
+		return stateOptions.DataAttributesLoadingPolicy
+	}
+
+	return stateOptions.WaitUntilApiDataAttributesLoadingPolicy
+}
```
longquanzheng, line 62: I think the expected behavior is to prefer more specific to override less specific  -- waitUntilDALoadingPolicy will override daLoadingPolicy, then DataObjectsLoadingPolicy, then nil

integ/workflow/wf_state_options_data_attributes_loading/routers.go
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/indeedeng/iwf/service"
+	"github.com/stretchr/testify/assert"
+	"log"
+	"net/http"
+)
+
+/**
+ * This test has four states
+ *
+ * State1: Creates all Data Attributes keys that will be used in this test
```
longquanzheng, line 16: Can you provide details on what does WaitUntil and Execute do? F.g. State1 doesn't do anything in WaitUntil.

integ/workflow/wf_state_options_data_attributes_loading/routers.go
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/indeedeng/iwf/gen/iwfidl"
+	"github.com/indeedeng/iwf/integ/workflow/common"
+	"github.com/indeedeng/iwf/service"
+	"github.com/stretchr/testify/assert"
+	"log"
+	"net/http"
+)
+
+/**
+ * This test has four states
```
longquanzheng, line 14: Nice! I planned to add documentation to all the exsting integ test workflow, and wondering what is the format. I think it's a great idea that just put it at the beginning.

Nit: `This test has four states` -> `This test workflow has four states, using REST controller to implement the workflow directly.`

integ/workflow/wf_state_options_data_attributes_loading/routers.go
```diff
@@ -0,0 +1,292 @@
...
+ *
+ * State1: Creates all Data Attributes keys that will be used in this test
+ * 		- da_wait_until1
+ * 		- da_execute1
+ *		- da_other_key
+ * State2:
+ * 		- Declares State Options containing WaitUntilApiDataAttributesLoadingPolicy
+ * 		- Asserts that (ApiV1WorkflowStateStart) WaitUntil method will load with expected Data Attributes
```
longquanzheng, line 22: also better to clarirfy where is the `Asserts`: in waitUntil or execute

service/common/compatibility/stateOptions.go
samuel27m: When I did this I tried copying from the existing fallback from `DataAttributesLoadingPolicy` to `HasDataObjectsLoadingPolicy`: 

```
        if stateOptions.HasDataObjectsLoadingPolicy() {
		return stateOptions.DataObjectsLoadingPolicy
	}
	return stateOptions.DataAttributesLoadingPolicy
```

But you're absolutely right, more specific should override less specific, I'll make the changes.
I will also implement a `integ` test that ensures that the more specific policy is the one being picked up if they're both specified.

Makefile
```diff
@@ -138,7 +138,7 @@ iwf-server:
 
 .PHONY: bins release clean
 
-idl-code-gen: #generate/refresh go clent code for idl, do this after update the idl file
+idl-code-gen: #generate/refresh go client code for idl, do this after update the idl file
```
longquanzheng, line 141: 👍

integ/workflow/wf_state_options_data_attributes_loading/routers.go
samuel27m: I did specify where the Asserts is: `- Asserts that (ApiV1WorkflowStateStart) **WaitUntil** method will load with expected Data Attributes`

But this probably means that it's not very readable. I'll make changes so that this reads better!

Makefile
samuel27m: Will do! Thanks!

integ/workflow/wf_state_options_data_attributes_loading/routers.go
samuel27m: Will change! Thanks!

Do you think that is best for the comment to be at the start of the routers file (where it currently is) or at the start of the test file (i.e. `wf_state_options_data_attributes_loading_test.go`)?

service/common/compatibility/stateOptions.go
samuel27m: This should be done now ✅

integ/workflow/wf_state_options_data_attributes_loading/routers.go
samuel27m: Done ✅

integ/workflow/wf_state_options_data_attributes_loading/routers.go
samuel27m: Done ✅

integ/workflow/wf_state_options_data_attributes_loading/routers.go
samuel27m: I made changes to the comment structure so that it is more readable. Let me know what you think, we can also brainstorm about it on a call if you prefer! 👍

Makefile
samuel27m: Done ✅

service/common/compatibility/stateOptions.go
```diff
@@ -45,14 +45,62 @@ func GetDecideApiRetryPolicy(stateOptions *iwfidl.WorkflowStateOptions) *iwfidl.
 	return stateOptions.ExecuteApiRetryPolicy
 }
 
-func GetDataObjectsLoadingPolicy(stateOptions *iwfidl.WorkflowStateOptions) *iwfidl.PersistenceLoadingPolicy {
+func GetWaitUntilApiDataObjectsLoadingPolicy(stateOptions *iwfidl.WorkflowStateOptions) *iwfidl.PersistenceLoadingPolicy {
```
longquanzheng, line 48: DataObject is the old name, let's rename it to the new one if possible (there are more we need to do in the repo)

longquanzheng (approved):

service/common/compatibility/stateOptions.go
samuel27m: I'll create a new issue for this 👍

PR #453, opened 2024-10-16
IWF-129: Enable executingStateIds SAs for states with waitUntil

### LINKED ISSUE #411: Enable `executingStateIds` system search attributes for only states that have waitUntil
The current `executingStateIds` is too expensive when setting the SA value for every state -- where most states don't have `waitUntil` and will get passed very quickly. 

Mostly, 99% the cases, people only need it when a state has a `waitUntil`. 

This will be very handy to find the states that are waiting on something. Knowing these states will help removing the code.

Ideally, the options to provide to client (via WorkflowConfig)
* Enable it for any state (even doesn't have waitUntil)
* Enable it only for state that has waitUntil (preferred as default)
* Disable (right now this is default for internal)

lwolczynski:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #411


lwolczynski:
I'm getting NDEs for:
- v2-basic.json
- v2-any-timer-signal.json
- v3-skip-start.json

Trying to dig into it. Let me know if you have any ideas @longquanzheng

service/interpreter/stateExecutionCounter.go
```diff
@@ -69,55 +69,106 @@ func (e *StateExecutionCounter) CreateNextExecutionId(stateId string) string {
...
 	e.totalCurrentlyExecutingCount += numOfNew
+
 	if needsUpdateSA {
-		return e.updateStateIdSearchAttribute()
+		return e.UpdateStateIdSearchAttribute()
 	}
 	return nil
 }
 
+func (e *StateExecutionCounter) IncreaseStateIdCurrentlyExecutingCounts(s iwfidl.StateMovement) bool {
```
longquanzheng, line 116: In Golang, capitalizing first letter make it "public", otherwise it's "private" within the package.
Here in iwf, it is just a convention that we only capitalizing the methods that are being called outside of this file. So if `IncreaseStateIdCurrentlyExecutingCounts` is only being used within this file (by other methods), we would keep it private. This is to make the code a little more readable (moving them into different packages will be very hard because Golang doesn't allow circular dependency acorss packages)

config/config.go
```diff
@@ -107,6 +107,7 @@ type (
 var DefaultWorkflowConfig = &iwfidl.WorkflowConfig{
 	ContinueAsNewThreshold:       iwfidl.PtrInt32(100),
 	DisableSystemSearchAttribute: iwfidl.PtrBool(true),
+	ExecutingStateIdMode:         iwfidl.PtrString("ENABLED_FOR_STATES_WITH_WAIT_UNTIL"),
 }
```
longquanzheng, line 111: I think we can just remove both and only keep `ContinueAsNewThreshold`. 
`ENABLED_FOR_STATES_WITH_WAIT_UNTIL` is the default behavior already, we don't have to pass it. It will be passed through workflow input, which will consume a some temporal storage 🤦 , for `ExecutingStateIdMode: "ENABLED_FOR_STATES_WITH_WAIT_UNTIL",` it will be 60 bytes. and if a workflow execution run for a month, it will be 60 bytes * month Lol. 

The reason that I added `DisableSystemSearchAttribute` is that the default behavior was different, and I didn't have time to change the default behavior using versioning.

service/interpreter/stateExecutionCounter.go
```diff
@@ -69,40 +69,94 @@ func (e *StateExecutionCounter) CreateNextExecutionId(stateId string) string {
...
+	config := e.configer.Get()
+
+	if e.globalVersioner.IsAfterVersionOfExecutingStateIdMode() {
+		switch mode := config.GetExecutingStateIdMode(); mode {
+		case "DISABLED":
+			return nil
+		case "ENABLED_FOR_ALL":
+			e.decreaseStateIdCurrentlyExecutingCounts(state)
+		default: // "ENABLED_FOR_STATES_WITH_WAIT_UNTIL" or nil or unrecognized enum value
+			if options.GetSkipWaitUntil() {
```
longquanzheng, line 142: We should use `skipStart := compatibility.GetSkipStartApi(&options)` like above for compatibity. Or just use the above resolved value.

service/interpreter/stateExecutionCounter.go
```diff
@@ -69,40 +69,94 @@ func (e *StateExecutionCounter) CreateNextExecutionId(stateId string) string {
...
+			switch mode := config.GetExecutingStateIdMode(); mode {
+			case "DISABLED":
+				// do nothing
+			case "ENABLED_FOR_ALL":
+				if e.increaseStateIdCurrentlyExecutingCounts(s) {
+					needsUpdateSA = true
+				}
+			default: // "ENABLED_FOR_STATES_WITH_WAIT_UNTIL" or nil or unrecognized enum value
+				options := s.GetStateOptions()
+				if !options.GetSkipWaitUntil() {
```
longquanzheng, line 94: similar here: `compatibility.GetSkipStartApi(&options)` 

Nit: it would be nice to rename to `compatibility.GetSkipWaitUntilApi(&options)`

service/interpreter/stateExecutionCounter.go
```diff
@@ -114,10 +168,7 @@ func (e *StateExecutionCounter) updateStateIdSearchAttribute() error {
 	for sid := range e.stateIdCurrentlyExecutingCounts {
 		executingStateIds = append(executingStateIds, sid)
 	}
-	config := e.configer.Get()
-	if config.GetDisableSystemSearchAttribute() {
-		return nil
-	}
```
longquanzheng, line 120: this took me a few minutes to realize that's it's not breaking determinsm! you are right, and I agreed that moving this check to upper makes it easier to read.

service/interpreter/stateExecutionCounter.go
```diff
@@ -69,55 +69,106 @@ func (e *StateExecutionCounter) CreateNextExecutionId(stateId string) string {
...
-		e.stateIdCurrentlyExecutingCounts[s.StateId]++
-		if e.stateIdCurrentlyExecutingCounts[s.StateId] == 1 {
-			// first time the stateId show up
-			needsUpdateSA = true
+
+		if e.globalVersioner.IsAfterVersionOfExecutingStateIdMode() {
+			switch mode := config.GetExecutingStateIdMode(); mode {
+			case "DISABLED":
+				// do nothing
+			case "ENABLED_FOR_ALL":
```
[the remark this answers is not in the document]
lwolczynski: `case "ENABLED_FOR_ALL", default:` doesn't seem to be the correct syntax

I changed it to:
```
		switch mode := config.GetExecutingStateIdMode(); mode {
		case iwfidl.DISABLED:
			return nil
		case iwfidl.ENABLED_FOR_ALL:
			e.decreaseStateIdCurrentlyExecutingCounts(state)
		case iwfidl.ENABLED_FOR_STATES_WITH_WAIT_UNTIL:
			fallthrough
		default:
			if options.GetSkipWaitUntil() {
				return nil
			} else {
				e.decreaseStateIdCurrentlyExecutingCounts(state)
			}
		}

```

service/interpreter/stateExecutionCounter.go
```diff
@@ -114,10 +168,7 @@ func (e *StateExecutionCounter) updateStateIdSearchAttribute() error {
 	for sid := range e.stateIdCurrentlyExecutingCounts {
 		executingStateIds = append(executingStateIds, sid)
 	}
-	config := e.configer.Get()
-	if config.GetDisableSystemSearchAttribute() {
-		return nil
-	}
+
 	if e.globalVersioner.IsAfterVersionOfOptimizedUpsertSearchAttribute() && len(executingStateIds) == 0 {
```
longquanzheng, line 176: This optimization is broken already for current default/ENABLED_FOR_ALL, and it won't work for ENABLED_FOR_STATES_WITH_WAIT_UNTIL. We need to fix it, and have to update `ClearExecutingStateIdsSearchAttributeFinally` as well.

To keep determinsm, we have to add a `IsBeforeVersionOfExecutingStateIdMode` in globalVersioner. 
Then here:
```
if e.globalVersioner.IsAfterVersionOfOptimizedUpsertSearchAttribute() && 
e.globalVersioner.IsBeforeVersionOfExecutingStateIdMode() &&
len(executingStateIds) == 0 {
    // explain why we stopped using this optimization
    return nil
}
```

Similarly in ClearExecutingStateIdsSearchAttributeFinally:
```
if e.globalVersioner.IsAfterVersionOfOptimizedUpsertSearchAttribute() && 
e.globalVersioner.IsBeforeVersionOfExecutingStateIdMode() {
   if if config.GetDisableSystemSearchAttribute() {	
		return
   }
   upsertSearchAttribute( []string{}) // clear the value.
}
```

Btw, you can also create a method in globalVersioner `IsAfterVersionOfBrokenOptimizedUpsertSearchAttributeAndBeforeVersionOfExecutingStateIdMode()` to combinet this check. Maybe that will look better.

longquanzheng: Btw, I realized that the correct way to optimize this UpsertSearchAttribute(which is a temporal action) is to check the next state like this: https://github.com/indeedeng/iwf/issues/454

We can also do it in next PR -- as long as we don't release the this PR to production, we don't have to create a new version

config/config.go
lwolczynski: Done

service/interpreter/stateExecutionCounter.go
lwolczynski: Done

service/interpreter/stateExecutionCounter.go
lwolczynski: Done

service/interpreter/stateExecutionCounter.go
lwolczynski: `IsBeforeVersionOfExecutingStateIdMode` is equal to `!IsAfterVersionOfExecutingStateIdMode`. Do we need this extra method?

The latter method is somehow malformed in your snippet. I changed it to:
```
func (e *StateExecutionCounter) ClearExecutingStateIdsSearchAttributeFinally() {
	config := e.configer.Get()

	if e.globalVersioner.IsAfterVersionOfExecutingStateIdMode() {
		if config.GetExecutingStateIdMode() == "DISABLED" {
			return
		}
	} else {
		if config.GetDisableSystemSearchAttribute() {
			return
		}
	}

	if e.globalVersioner.IsAfterVersionOfOptimizedUpsertSearchAttribute() && !e.globalVersioner.IsAfterVersionOfExecutingStateIdMode() && e.totalCurrentlyExecutingCount == 0 {
		err := e.provider.UpsertSearchAttributes(e.ctx, map[string]interface{}{
			service.SearchAttributeExecutingStateIds: []string{},
		})
		if err != nil {
			e.provider.GetLogger(e.ctx).Error("error for upsert SearchAttributeExecutingStateIds", err)
		}
	}
}
```

I believe this ^^ is correct.

And I'd rather work on #454 in a separate branch. I can rebase off of this one and merge #454 into here before merging to main if we want to release other changes.

integ/wait_until_search_attributes_test.go
```diff
@@ -0,0 +1,115 @@
...
+			ExecutingStateIdMode: ptr.Any(iwfidl.ENABLED_FOR_ALL),
+		})
+		smallWaitForFastTest()
+	}
+
+	// TODO: Rethink how this can be tested
+	// for i := 0; i < *repeatIntegTest; i++ {
+	// doTestWaitUntilSearchAttributes(t, service.BackendTypeTemporal, nil) // defaults to ExecutingStateIdMode: ENABLED_FOR_STATES_WITH_WAIT_UNTIL
+	// smallWaitForFastTest()
+	// }
```
lwolczynski, line 39: The last thing to do here is adding a test for the `skipWaitUntil` case

service/interpreter/stateExecutionCounter.go
longquanzheng: > IsBeforeVersionOfExecutingStateIdMode is equal to !IsAfterVersionOfExecutingStateIdMode. Do we need an extra method for this check?

You are right, we don't have to. Sometimes I just prefer to not use negative statement/expression (reasons are like https://www.obsidiansecurity.com/blog/why-to-stop-writing-negative-code/  ) 

But no much difference and it's not always the case.

service/interpreter/stateExecutionCounter.go
```diff
@@ -133,10 +188,18 @@ func (e *StateExecutionCounter) updateStateIdSearchAttribute() error {
...
+
+	if e.globalVersioner.IsAfterVersionOfExecutingStateIdMode() {
+		if config.GetExecutingStateIdMode() == "DISABLED" {
+			return
+		}
+	} else {
+		if config.GetDisableSystemSearchAttribute() {
+			return
+		}
 	}
```
longquanzheng, line 200: I may miss something, do we need this block? If the version is not after, the next block will not do anything ( `!e.globalVersioner.IsAfterVersionOfExecutingStateIdMode()` )

It seems like we just need to keep
```
if config.GetDisableSystemSearchAttribute() {
		return
```

longquanzheng: btw, you can consider this as a Nit comment. It is not wrong, but just not necessary :D

integ/wait_until_search_attributes_test.go
longquanzheng: 👍 
Potentially we can just use a single workflow with two states to test it:
```
S1: skipWaitUntil, execute will just sleep for 5 seconds, then go to S2
S2: waitUntil for a signalA, execute will then go to S3
S3: waitUntil for a signalB, execute will then go to S3
```
S1 will have a 5 seconds window to assert for the SA.
S2/S3 can let us assert on the SA and then send the signal (and verify the previous one is gone)

This can be testing against different modes.

service/interpreter/stateExecutionCounter.go
longquanzheng: >And I'd rather work on https://github.com/indeedeng/iwf/issues/454 in a separate branch. I can rebase off of this one and merge https://github.com/indeedeng/iwf/issues/454 into here before merging to main if we want to release other changes. 

+1 
Sounds good. btw maybe not add replay tests for this PR . Otherwise you will have to do it again if we want to reuse the same version (because that optimization is a non-determinstic change)

integ/wait_until_search_attributes_test.go
```diff
@@ -0,0 +1,115 @@
...
+	case iwfidl.ENABLED_FOR_STATES_WITH_WAIT_UNTIL:
+		assertSearch(fmt.Sprintf("WorkflowId='%v'", wfId), 1, apiClient, assertions)
+		// TODO: Add search attribute assert
+	case iwfidl.DISABLED:
+		assertSearch(fmt.Sprintf("WorkflowId='%v'", wfId), 1, apiClient, assertions)
+		assertSearch(fmt.Sprintf("WorkflowId='%v' AND %v='%v'", wfId, wait_until_search_attributes.TestSearchAttributeExecutingStateIdsKey, wait_until_search_attributes.State2), 0, apiClient, assertions)
+	}
+
+	time.Sleep(time.Second * 5) // wait for a few seconds so that timer is ready to be skipped
+	req3 := apiClient.DefaultApi.ApiV1WorkflowTimerSkipPost(context.Background())
```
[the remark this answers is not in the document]
lwolczynski: Simplified this test and removed the command entirely

lwolczynski:
Add comments addressed. Moving on to #454

longquanzheng (approved):

PR #460, opened 2024-10-21
IWF-208: Optimize executingStateIdMode for state transition

### LINKED ISSUE #454: Optimize executingStateIdMode of ENABLED_FOR_STATES_WITH_WAIT_UNTIL
Check if the next state is also wait until, then skip a upsertSearchAttribute as clearing up for the current one(because the next one will be upserting )

Note: this will need a new version after https://github.com/indeedeng/iwf/pull/453 if not built within

lwolczynski:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #454

### DIFF
--- a/integ/wait_until_search_attributes_optimization_test.go
+++ b/integ/wait_until_search_attributes_optimization_test.go
@@ -69,7 +69,7 @@ func doTestWaitUntilHistoryCompleted(
 	wfReq := iwfidl.WorkflowStartRequest{
 		WorkflowId:             wfId,
 		IwfWorkflowType:        wait_until_search_attributes_optimization.WorkflowType,
-		WorkflowTimeoutSeconds: 10,
+		WorkflowTimeoutSeconds: 15,
 		IwfWorkerUrl:           "http://localhost:" + testWorkflowServerPort,
 		StartStateId:           ptr.Any(wait_until_search_attributes_optimization.State1),
 		WorkflowStartOptions: &iwfidl.WorkflowStartOptions{
@@ -129,20 +129,30 @@ func doTestWaitUntilHistoryCompleted(
 
 	switch mode := config.GetExecutingStateIdMode(); mode {
 	case iwfidl.ENABLED_FOR_ALL:
-		assertions.Equal(6, len(upsertSAEvents))
+		assertions.Equal(10, len(upsertSAEvents))
 		assertions.Equal("[\"S1\"]", historyEventSAs(upsertSAEvents[0]))
 		assertions.Equal("[\"S2\"]", historyEventSAs(upsertSAEvents[1]))
-		assertions.Equal("[\"S3\",\"S4\"]", historyEventSAs(upsertSAEvents[2]))
-		assertions.Equal("[\"S4\"]", historyEventSAs(upsertSAEvents[3]))
-		assertions.Equal("[\"S5\"]", historyEventSAs(upsertSAEvents[4]))
-		assertions.Equal("null", historyEventSAs(upsertSAEvents[5]))
+		assertions.Equal("[\"S2\",\"S3\"]", historyEventSAs(upsertSAEvents[2]))
+		assertions.Equal("[\"S4\",\"S3\"]", historyEventSAs(upsertSAEvents[3]))
+		assertions.Equal("[\"S5\",\"S3\"]", historyEventSAs(upsertSAEvents[4]))
+		assertions.Equal("[\"S6\",\"S3\",\"S7\"]", historyEventSAs(upsertSAEvents[5]))
+		assertions.Equal("[\"S6\",\"S3\"]", historyEventSAs(upsertSAEvents[6]))
+		// TODO: This is unexpected; should not upsert the same SAs -- happens after "_SYS_GRACEFUL_COMPLETING_WORKFLOW"
+		assertions.Equal("[\"S3\"]", historyEventSAs(upsertSAEvents[7]))
+		assertions.Equal("[\"S3\"]", historyEventSAs(upsertSAEvents[8]))
+		assertions.Equal("null", historyEventSAs(upsertSAEvents[9]))
 	case iwfidl.ENABLED_FOR_STATES_WITH_WAIT_UNTIL:
-		assertions.Equal(5, len(upsertSAEvents))
+		assertions.Equal(9, len(upsertSAEvents))
 		assertions.Equal("[\"S1\"]", historyEventSAs(upsertSAEvents[0]))
 		assertions.Equal("[\"S2\"]", historyEventSAs(upsertSAEvents[1]))
-		assertions.Equal("[\"S3\",\"S4\"]", historyEventSAs(upsertSAEvents[2]))
-		assertions.Equal("[\"S4\"]", historyEventSAs(upsertSAEvents[3]))
-		assertions.Equal("null", historyEventSAs(upsertSAEvents[4]))
+		assertions.Equal("[\"S2\",\"S3\"]", historyEventSAs(upsertSAEvents[2]))
+		assertions.Equal("[\"S4\",\"S3\"]", historyEventSAs(upsertSAEvents[3]))
+		assertions.Equal("[\"S3\"]", historyEventSAs(upsertSAEvents[4]))
+		assertions.Equal("[\"S6\",\"S3\"]", historyEventSAs(upsertSAEvents[5]))
+		// TODO: This is unexpected; should not upsert the same SAs -- happens after "_SYS_GRACEFUL_COMPLETING_WORKFLOW"
+		assertions.Equal("[\"S3\"]", historyEventSAs(upsertSAEvents[6]))
+		assertions.Equal("[\"S3\"]", historyEventSAs(upsertSAEvents[7]))
+		assertions.Equal("null", historyEventSAs(upsertSAEvents[8]))
 	case iwfidl.DISABLED:
 		assertions.Equal(0, len(upsertSAEvents))
 	}
--- a/integ/wait_until_search_attributes_optimization_test.go
+++ b/integ/wait_until_search_attributes_optimization_test.go
@@ -8,7 +8,9 @@ import (
 	"go.temporal.io/api/enums/v1"
 	history "go.temporal.io/api/history/v1"
 	"go.temporal.io/api/workflowservice/v1"
+	"slices"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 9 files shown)


service/interpreter/stateExecutionCounter.go
```diff
@@ -180,6 +191,23 @@ func (e *StateExecutionCounter) updateStateIdSearchAttribute() error {
 		// see workflowImpl.go to call ClearExecutingStateIdsSearchAttributeFinally at the end
 		return nil
 	}
+
+	if stateTransition != nil {
+		// UpsertSearchAttributes should be only invoked if when the next states do not skip waitUntil
+		shouldSkipUpsertingSAs := true
+		for _, s := range stateTransition.next {
+			if !s.StateOptions.GetSkipWaitUntil() {
+				shouldSkipUpsertingSAs = false
```
mixydubs, line 200: I think perhaps I'm missing some context or understanding of what we want to achieve here - what happens when the first StateMovement in the next array skips waitUntil, but subsequent ones do not?  If I'm reading this correctly, with this change if any next StateMovement does not have a waitUntil, the `SearchAttributeExecutingStateIds` will not be updated...

mixydubs: hmm... I have that wrong don't I... if any of the next states do have a waitUntil (i.e. they don't skip wait until) then do continue on to UpsertSearchAttributes??

lwolczynski: Right, if any of the next states don't skip waitUntil, then the SAs will be upserted. Also, if I'm not mistaken, nextStates are the states that are executed in the state transition, example:
current = A
next = [B, C, D]
means that B, C, and D are executed at the same time after A is completed.

service/interpreter/stateExecutionCounter.go
```diff
@@ -135,27 +137,61 @@ func (e *StateExecutionCounter) MarkStateExecutionCompleted(state iwfidl.StateMo
...
 		} else {
-			e.decreaseStateIdCurrentlyExecutingCounts(state)
+			e.decreaseStateIdCurrentlyExecutingCounts(currentState)
 		}
 	}
 
 	return e.updateStateIdSearchAttribute()
 }
 
+func determineIfShouldSkipUpsert(currentState iwfidl.StateMovement, nextStates []iwfidl.StateMovement) bool {
```
longquanzheng, line 169: I thought the signature would have pass `ExecutingStateIdMode`.
I think the currentState can be omitted. 

```
func determineIfShouldSkipUpsert(mode ExecutingStateIdMode, nextStates ) bool{
  // based on the mode
  // For ENABLED_FOR_ALL, return true as long as any of next states is going to a next state(not closing state)
  // For ENABLED_FOR_STATES_WITH_WAIT_UNTIL, in addition to above, also check the state has waitUntil.
}
```
StateMovement may not go to any state, because we are doing a hack in iwf (I really regret that) 😨 
We use some special StateIDs to prepresent gracefulComplete, deadEnd etc. See https://github.com/indeedeng/iwf/blob/main/service/const.go#L68
And how it's being used in https://github.com/indeedeng/iwf/blob/main/service/interpreter/workflowImpl.go#L373

lwolczynski: > I thought the signature would have pass ExecutingStateIdMode.
I'm not passing it as I only invoke determineIfShouldSkipUpsert when needed based on `ExecutingStateIdMode`

currentState is needed if we want to determine when a state is looping back, I believe

longquanzheng: > currentState is needed if we want to determine when a state is looping back, I believe
for that case, I belive what you have in  `updateStateIdSearchAttribute` baesd on comparing with GetSearchAttributes will cover it?

longquanzheng: Basically, if `MarkStateExecutionCompleted` not delete the SA, next time when `MarkStateExecutingIfNotYet`->`updateStateIdSearchAttribute ` will then skip it because the SA is still there

service/interpreter/stateExecutionCounter.go
```diff
@@ -84,15 +97,17 @@ func (e *StateExecutionCounter) MarkStateIdExecutingIfNotYet(stateReqs []StateRe
 			case iwfidl.DISABLED:
 				// do nothing
 			case iwfidl.ENABLED_FOR_ALL:
-				if e.increaseStateIdCurrentlyExecutingCounts(s) {
+				e.increaseStateIdCurrentlyExecutingCounts(s)
+				if !slices.Contains(currentSAs.StringArrayValue, s.StateId) {
```
longquanzheng, line 101: After reflecting into this for a while, I found issues with this comparison. Just checking containing is probably not correct.

Consider this case (assuming using ENABLED_FOR_ALL so that we don't care about the waitUntil):
* S1->S2+S1
* S2->S1
So there are 3 S1 executions. Assuming the first S1-1 completed, then goes to S2-1 and S1-2. But S1-2 didn't complete yet, then S2-1 completed and decided to go to S1-3.
Now S1-3 will call this `MarkStateIdExecutingIfNotYet`:
* Because of the optimization in `MarkStateExecutionCompleted`, `S2` is not deleted and still in the search attribute. So the SA contains both: [S1, S2]. (Assuming you are implementing the optimization as we discussed today -- skip deleting the stateID from SA if there is a next state going to refresh the whole SA)
* Now for S1-3 calling `MarkStateIdExecutingIfNotYet`, it will see the SA contains S1 already, but it also contains S2. Therefore, it should still refresh the SA based on `stateIdCurrentlyExecutingCounts` (Note: it may be a lot better to rename `updateStateIdSearchAttribute` -> `refreshIwfExecutingStateIdSearchAttribute` to be more clear.

I remember you were doing the full comparison when you put it in the `updateStateIdSearchAttribute`, not sure why you changed your mind 🤣 

I think you may still run into NDE if putting it into that method, but instead, maybe you can create another method `checkIfNeedToRefreshIwfExecutingStateIdSearchAttribute` before calling `refreshIwfExecutingStateIdSearchAttribute`.

service/interpreter/stateExecutionCounter.go
```diff
@@ -109,8 +109,37 @@ func (e *StateExecutionCounter) MarkStateIdExecutingIfNotYet(stateReqs []StateRe
...
+			e.provider.GetLogger(e.ctx).Error("error for GetSearchAttributes", err)
+		}
+
+		var currentSAsValues []string
+
+		currentSAs, ok := sas[service.SearchAttributeExecutingStateIds]
+		if ok {
+			currentSAsValues = currentSAs.StringArrayValue
+		} else {
+			e.provider.GetLogger(e.ctx).Error("search attribute IwfExecutingStateIds is not found", err)
```
longquanzheng, line 128: I feel like it's epxected to have empty/nil value to start with, so we probably just ignore this case, or just assign a empty slice so that `DeepEqual` can return `true` when comparing with another empty slice

service/interpreter/stateExecutionCounter.go
```diff
@@ -135,25 +164,56 @@ func (e *StateExecutionCounter) MarkStateExecutionCompleted(state iwfidl.StateMo
 		case iwfidl.DISABLED:
 			return nil
 		case iwfidl.ENABLED_FOR_ALL:
-			e.decreaseStateIdCurrentlyExecutingCounts(state)
+			e.decreaseStateIdCurrentlyExecutingCounts(currentState)
+			shouldSkipUpsert := determineIfShouldSkipRefreshOnCompleted(nextStates, true)
```
longquanzheng, line 168: Nit: `shouldSkipUpsert` -> `shouldSkipRefresh`. I feel like one of the confusion is that this is not really upserting new stateID to the SA, but deleting statID by refreshing the whole map (keys of stateIdCurrentlyExecutingCounts)

service/interpreter/stateExecutionCounter.go
```diff
@@ -135,25 +164,56 @@ func (e *StateExecutionCounter) MarkStateExecutionCompleted(state iwfidl.StateMo
...
+			nonClosingNextStates = append(nonClosingNextStates, s)
+		}
+	}
+	if enabledForAll {
+		if len(nonClosingNextStates) > 0 {
+			return true
+		}
+	} else {
+		for _, s := range nonClosingNextStates {
+			if !s.StateOptions.GetSkipWaitUntil() {
```
longquanzheng, line 210: It should be `!compatibility.GetSkipWaitUntilApi(&options)` like you did in https://github.com/indeedeng/iwf/pull/453 

Btw, this double negative i hard to follow. maybe you can make it  `compatibility.HasWaitUntilApi(&options)` which will make it more clear in this statement -- when any of the next states has waitUntil, return true to skip refresh (because the next state will refresh immediately)

longquanzheng:
Looks good now. Thanks for the hard work!

service/interpreter/stateExecutionCounter.go
lwolczynski: Do we need `!compatibility.GetSkipWaitUntilApi(&options)`? This check is only for the newest version so I thought using `s.StateOptions.GetSkipWaitUntil()` was safe

lwolczynski: I changed it, but would like to understand better why not use `s.StateOptions.GetSkipWaitUntil()` in this case. Unless it's only a safety measure

service/interpreter/stateExecutionCounter.go
lwolczynski: Done

integ/wait_until_search_attributes_optimization_test.go
```diff
@@ -0,0 +1,154 @@
...
+		assertions.Equal("[\"S4\"]", historyEventSAs(upsertSAEvents[3]))
+		assertions.Equal("[\"S5\"]", historyEventSAs(upsertSAEvents[4]))
+		assertions.Equal("null", historyEventSAs(upsertSAEvents[5]))
+	case iwfidl.ENABLED_FOR_STATES_WITH_WAIT_UNTIL:
+		assertions.Equal(5, len(upsertSAEvents))
+		assertions.Equal("[\"S1\"]", historyEventSAs(upsertSAEvents[0]))
+		assertions.Equal("[\"S2\"]", historyEventSAs(upsertSAEvents[1]))
+		assertions.Equal("[\"S3\",\"S4\"]", historyEventSAs(upsertSAEvents[2]))
+		assertions.Equal("[\"S4\"]", historyEventSAs(upsertSAEvents[3]))
+		assertions.Equal("null", historyEventSAs(upsertSAEvents[4]))
```
lwolczynski, line 145: I think I tested all the usecases we talked about @longquanzheng:
- S1 loops to itself 5 times
- S3, S4 -> S4 is the one where `slices.Contains` check would fail

service/interpreter/stateExecutionCounter.go
longquanzheng: > This check is only for the newest version so I thought using s.StateOptions.GetSkipWaitUntil() was safe

Oh, you are right. It's not needed because of the versioning

integ/wait_until_search_attributes_optimization_test.go
lwolczynski: On the second thought, I'll add one more. S3, S4 -> S4 **-> S4, S5, ->** S5 -> END

integ/wait_until_search_attributes_optimization_test.go
```diff
@@ -0,0 +1,164 @@
...
+		assertions.Equal(9, len(upsertSAEvents))
+		assertions.Equal("[\"S1\"]", historyEventSAs(upsertSAEvents[0]))
+		assertions.Equal("[\"S2\"]", historyEventSAs(upsertSAEvents[1]))
+		assertions.Equal("[\"S2\",\"S3\"]", historyEventSAs(upsertSAEvents[2]))
+		assertions.Equal("[\"S4\",\"S3\"]", historyEventSAs(upsertSAEvents[3]))
+		assertions.Equal("[\"S3\"]", historyEventSAs(upsertSAEvents[4]))
+		assertions.Equal("[\"S6\",\"S3\"]", historyEventSAs(upsertSAEvents[5]))
+		// TODO: This is unexpected; should not upsert the same SAs -- happens after "_SYS_GRACEFUL_COMPLETING_WORKFLOW"
+		assertions.Equal("[\"S3\"]", historyEventSAs(upsertSAEvents[6]))
+		assertions.Equal("[\"S3\"]", historyEventSAs(upsertSAEvents[7]))
```
lwolczynski, line 154: TODO: Troubleshoot this behavior. Should not upsert as there is no change

service/interpreter/stateExecutionCounter.go
```diff
@@ -109,8 +109,35 @@ func (e *StateExecutionCounter) MarkStateIdExecutingIfNotYet(stateReqs []StateRe
...
+		if ok {
+			currentSAsValues = currentSAs.StringArrayValue
+		}
+
+		var executingStateIds []string
+		for sid := range e.stateIdCurrentlyExecutingCounts {
+			executingStateIds = append(executingStateIds, sid)
+		}
+
+		if reflect.DeepEqual(currentSAsValues, executingStateIds) {
```
longquanzheng, line 134: Actually, I think we should use something like elementMatch to compare ? Because the order could be different https://stackoverflow.com/questions/36000487/check-for-equality-on-slices-without-order

lwolczynski: I sort the slice (line 170)

lwolczynski: But I agree doing elementMatch would be easier. I learned about it after I added sorting... 🙃

integ/wait_until_search_attributes_optimization_test.go
```diff
@@ -0,0 +1,172 @@
...
+		assertions.Equal([]string{"S2"}, historyEventSAs(upsertSAEvents[1]))
+		assertions.Equal([]string{"S2", "S3"}, historyEventSAs(upsertSAEvents[2]))
+		assertions.Equal([]string{"S3", "S4"}, historyEventSAs(upsertSAEvents[3]))
+		assertions.Equal([]string{"S3", "S5"}, historyEventSAs(upsertSAEvents[4]))
+		assertions.Equal([]string{"S3", "S6", "S7"}, historyEventSAs(upsertSAEvents[5]))
+		assertions.Equal([]string{"S3", "S6"}, historyEventSAs(upsertSAEvents[6]))
+		// TODO: This is unexpected; should not upsert the same SAs -- happens after "_SYS_GRACEFUL_COMPLETING_WORKFLOW"
+		assertions.Equal([]string{"S3"}, historyEventSAs(upsertSAEvents[7]))
+		assertions.Equal([]string{"S3"}, historyEventSAs(upsertSAEvents[8]))
+		assertions.Equal([]string{"null"}, historyEventSAs(upsertSAEvents[9]))
```
lwolczynski, line 145: Technically we should not upsert "S3" twice. Seems that we always upsert on `_SYS_GRACEFUL_COMPLETING_WORKFLOW`

- SAs -> "S3", "S6", "S7"
- S7-1 -> _SYS_GRACEFUL_COMPLETING_WORKFLOW
- SAs -> "S3", "S6"
- S6-1 -> _SYS_GRACEFUL_COMPLETING_WORKFLOW
- SAs -> "S3"
- S3-1 -> _SYS_GRACEFUL_COMPLETING_WORKFLOW
- SAs -> "S3"
- S3-2 -> _SYS_GRACEFUL_COMPLETING_WORKFLOW
- SAs -> null

service/interpreter/stateExecutionCounter.go
longquanzheng: Nice. Sorry I didn’t read your latest code yet. I just thinking about it and wondering why you had NDE yesterday. This could be the reasons. Especially doing it in the update/refresh method for both MarkExecuting and MarkCompleted, it increases a lot the likelihood of NDE because of the randomness

longquanzheng: Have you push the latest code? I don’t see the sorting ( do you mean sorting the keys of the map, right?)

integ/wait_until_search_attributes_optimization_test.go
longquanzheng: Is that because of we miss an existing optimization— on MarkCompletion, we should only refresh when any stateID reaches to zero (needs to delete)

integ/wait_until_search_attributes_optimization_test.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func historyEventSAs(e *history.HistoryEvent) []string {
+	attrs := e.GetAttributes().(*history.HistoryEvent_UpsertWorkflowSearchAttributesEventAttributes)
+	data := string(attrs.UpsertWorkflowSearchAttributesEventAttributes.GetSearchAttributes().GetIndexedFields()[service.SearchAttributeExecutingStateIds].GetData())
+	data = strings.ReplaceAll(data, "[", "")
+	data = strings.ReplaceAll(data, "]", "")
+	data = strings.ReplaceAll(data, "\"", "")
+	dataSlice := strings.Split(data, ",")
+	slices.Sort(dataSlice)
```
lwolczynski, line 170: @longquanzheng

service/interpreter/stateExecutionCounter.go
lwolczynski: https://github.com/indeedeng/iwf/pull/460#pullrequestreview-2391203570

lwolczynski: Oh, I see the confusion. I was focusing on the tests... my bad. Yes, I should add sorting or elementMatch here.

lwolczynski: Added now

integ/wait_until_search_attributes_optimization_test.go
lwolczynski: Okay, this should be the last thing to tackle. I'll take care of it in the morning

longquanzheng: Latest thought: If the NDE is actually because of sorting and we can fix it and then move the logic back to the refresh method, then this should not be an issue and we don’t need to check if a StateId count reaches zero

service/interpreter/stateExecutionCounter.go
longquanzheng: Now with this sorting, can you move the whole optimization back to `refreshIwfExecutingStateIdSearchAttribute` again? I think if this solves the NDE, it will be also help the issue of https://github.com/indeedeng/iwf/pull/460#discussion_r1814203277

integ/wait_until_search_attributes_optimization_test.go
lwolczynski: Fixed this issue now and tests are passing as expected. We can try moving back to the refresh method

longquanzheng (approved):

PR #468, opened 2024-10-28
IWF-119: Options to ignore already started error

### LINKED ISSUE #312: Options to ignore already started error
It's been a common pattern to ignore AlreadyStartedError.

See example in Temporal Golang SDK's [StartWorkflowOptions](https://github.com/temporalio/sdk-go/blob/master/client/client.go#L523)

Implementation idea:
* Option 1: pass in the ignore as a flag in the start request. So that the server will respond 200 status. (preferred)
* Option 2: Alternatively, doing it only in SDK is also fine. But server will respond 400 status today, which will look a bit annoying.

Another big reason for option 1 is that we can support passing in a idempotent key(uuid) to dedup(already supported in cadence temporal). Only throw error if the key not match. 

This dedup key is super useful for case like parent starts child workflows from a channel message. Where msg can be duplicated. It can use state execution id as dedup key.

lwolczynski:
### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related Issue
Closes #312


service/api/service.go
```diff
@@ -119,6 +133,11 @@ func (s *serviceImpl) ApiV1WorkflowStartPost(
 				workflowOptions.Memo[da.GetKey()] = da.GetValue()
 			}
 		}
+		if requestId != nil {
+			workflowOptions.Memo[service.WorkflowAlreadyStartedRequestId] = iwfidl.EncodedObject{
+				Data: requestId,
+			}
```
lwolczynski, line 139: @longquanzheng when saving `WorkflowAlreadyStartedRequestId` to memo, should it be an EncodedObject or a simple string?

service/api/service.go
```diff
@@ -134,18 +152,37 @@ func (s *serviceImpl) ApiV1WorkflowStartPost(
...
 
 	runId, err := s.client.StartInterpreterWorkflow(ctx, workflowOptions, input)
 	if err != nil {
-		return nil, s.handleError(err, WorkflowStartApiPath, req.GetWorkflowId())
-	}
+		shouldHandleError := true
 
-	s.logger.Info("Started workflow", tag.WorkflowID(req.WorkflowId), tag.WorkflowRunID(runId))
+		if s.client.IsWorkflowAlreadyStartedError(err) && ignoreAlreadyStartedError {
+			response, descErr := s.client.DescribeWorkflowExecution(ctx, req.GetWorkflowId(), "", nil)
```
longquanzheng, line 166: I think we can check `requestId` before calling this API. If `requestId` is not provided, we can simply return 200 directly.

service/api/service.go
```diff
@@ -134,18 +152,37 @@ func (s *serviceImpl) ApiV1WorkflowStartPost(
...
+			if descErr != nil {
+				return nil, s.handleError(err, WorkflowStartApiPath, req.WorkflowId)
+			}
+
+			runId = response.RunId
```
longquanzheng, line 171: This shouldn't be different. I think we should pass the runId that returned from `StartInterpreterWorkflow` into `DescribeWorkflowExecution`. This is to be more accurate and avoid some edge cases from happening.

service/api/service.go
```diff
@@ -134,18 +152,37 @@ func (s *serviceImpl) ApiV1WorkflowStartPost(
...
 		WaitForCompletionStateExecutionIds: req.GetWaitForCompletionStateExecutionIds(),
 		WaitForCompletionStateIds:          req.GetWaitForCompletionStateIds(),
 		OmitVersionMarker:                  s.config.Api.OptimizedVersioning,
 	}
 
 	runId, err := s.client.StartInterpreterWorkflow(ctx, workflowOptions, input)
 	if err != nil {
-		return nil, s.handleError(err, WorkflowStartApiPath, req.GetWorkflowId())
-	}
+		shouldHandleError := true
```
longquanzheng, line 163: Nit: shouldReturnError

integ/wf_ignore_already_started_test.go
```diff
@@ -0,0 +1,150 @@
...
+	if errorExpected {
+		apiErr, ok := err.(*iwfidl.GenericOpenAPIError)
+		if !ok {
+			log.Fatalf("Should fail to invoke start api %v", err)
+		}
+		errResp, ok := apiErr.Model().(iwfidl.ErrorResponse)
+		if !ok {
+			log.Fatalf("should be error response")
+		}
+		assertions.Equal(iwfidl.WORKFLOW_ALREADY_STARTED_SUB_STATUS, errResp.GetSubStatus())
```
longquanzheng, line 131: nit: we should also assert on statusCode to be 400

longquanzheng (commented):
almost looks good! great job!

service/api/service.go
lwolczynski: Sure. Didn't know that that error would return `runId`. Addressed

service/api/service.go
lwolczynski: Addressed

service/api/service.go
lwolczynski: Addressed

integ/wf_ignore_already_started_test.go
lwolczynski: Addressed

longquanzheng (approved):

PR #473, opened 2024-10-30
IWF-232: Add disabling sticky cache option

### LINKED ISSUE #466: Run integ tests with sticky cache disabled to ensure determinism
Read through Go though the Temporal sticky cache [documentation](https://community.temporal.io/t/is-there-any-documentation-for-sticky-execution/4073) 

Probably use a command line parameter and run as a different action other the integ test time could double

lwolczynski:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #466

### DIFF
--- a/.github/workflows/ci-cadence-integ-test-disable-sticky.yml
+++ b/.github/workflows/ci-cadence-integ-test-disable-sticky.yml
@@ -9,12 +9,18 @@ jobs:
   tests:
     name: "Integration testing"
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        test-subset:
+          - "a-m"
+          - "n-z"
     steps:
       - uses: actions/checkout@v3
       - name: "Set up cadence environment"
         run: docker compose -f docker-compose/ci-cadence-dependencies.yml up -d
       - name: "Test against cadence"
-        run: make ci-cadence-integ-test
+        run: make ci-cadence-integ-test-disable-sticky startsWith=${{ matrix['test-subset'] }}
       - name: Dump docker logs
         if: always()
         uses: jwalton/gh-docker-logs@v2
--- a/.github/workflows/ci-cadence-integ-test-disable-sticky.yml
+++ b/.github/workflows/ci-cadence-integ-test-disable-sticky.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   tests:
-    name: "Integration testing"
+    name: "Integration testing with sticky cache disabled"
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
--- a/.github/workflows/ci-temporal-integ-test-disable-sticky.yml
+++ b/.github/workflows/ci-temporal-integ-test-disable-sticky.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   tests:
-    name: "Integration testing"
+    name: "Integration testing with sticky cache disabled"
     runs-on: ubuntu-latest
     permissions:
       # Give the default GITHUB_TOKEN write permission to commit and push the
--- a/.github/workflows/ci-temporal-integ-test-disable-sticky.yml
+++ b/.github/workflows/ci-temporal-integ-test-disable-sticky.yml
@@ -13,12 +13,19 @@ jobs:
       # Give the default GITHUB_TOKEN write permission to commit and push the
       # added or changed files to the repository.
       contents: write
+    strategy:
+      fail-fast: false
+      matrix:
+        test-subset:
+          - "a-m"
+          - "n-z"
+
     steps:
       - uses: actions/checkout@v3
       - name: "Set up temporal environment"
         run: docker compose -f docker-compose/ci-temporal-dependencies.yml up -d
       - name: "Test against temporal"
-        run: make ci-temporal-integ-test
+        run: make ci-temporal-integ-test-disable-sticky startsWith=${{ matrix['test-subset'] }}
       - name: Dump docker logs
         if: always()
         uses: jwalton/gh-docker-logs@v2
--- a/.github/workflows/ci-temporal-integ-test.yml
+++ b/.github/workflows/ci-temporal-integ-test.yml
@@ -13,12 +13,19 @@ jobs:
       # Give the default GITHUB_TOKEN write permission to commit and push the
       # added or changed files to the repository.
       contents: write
+    strategy:
+      fail-fast: false
+      matrix:
+        test-subset:
+          - "a-m"
+          - "n-z"
+
     steps:
       - uses: actions/checkout@v3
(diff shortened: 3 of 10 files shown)


service/interpreter/cadence/worker.go
```diff
@@ -19,6 +19,10 @@ type InterpreterWorker struct {
 	tasklist  string
 }
 
+type StartOptions struct {
+	DisableStickyCache bool
```
mixydubs, line 23: I know we have a specific use case for this start option to disable sticky cache, but I'm wondering if we should add a warning comment about the performance impacts of doing this for other reasons. (My understanding is that this would have a  not insignificant impact on performance in a non-test/real-world use case. Is that right?)

service/interpreter/cadence/worker.go
```diff
@@ -37,12 +41,17 @@ func (iw *InterpreterWorker) Close() {
 	iw.worker.Stop()
 }
 
-func (iw *InterpreterWorker) Start() {
+func (iw *InterpreterWorker) Start(startOptions StartOptions) {
```
mixydubs, line 44: Could we keep the original `Start` function here and have a second `StartWithOptions`  so that the `launchTemporalService` in `iwf.go` does not need to change?  The `Start` function could just set the default StartOptions and call StartWithOptions - I think that's cleaner overall but in particular keeps launchTemporalService clean

Unless we think we have a use case where launchTemporalService would actually be calling start with different options based on runtime configuration values??

lwolczynski: That makes sense to me. Definitely, we can keep `Start` as it was and add `StartWithOptions`. I like the idea, thanks for this suggestion!

.github/workflows/ci-temporal-integ-test.yml
```diff
@@ -13,12 +13,19 @@ jobs:
       # Give the default GITHUB_TOKEN write permission to commit and push the
       # added or changed files to the repository.
       contents: write
+    strategy:
```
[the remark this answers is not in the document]
lwolczynski: Got it. I thought we would only do tests with no sticky cache, but it makes sense to keep both scenarios 👍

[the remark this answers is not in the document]
lwolczynski: Done

service/interpreter/temporal/worker.go
```diff
@@ -36,6 +41,15 @@ func (iw *InterpreterWorker) Close() {
...
 func (iw *InterpreterWorker) Start() {
+	var options StartOptions
+
+	// default options
+	options.DisableStickyCache = false
+
+	iw.StartWithOptions(options)
+}
+
+func (iw *InterpreterWorker) StartWithOptions(startOptions StartOptions) {
```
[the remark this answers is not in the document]
lwolczynski: Done

.github/workflows/ci-cadence-integ-test-disable-sticky.yml
```diff
@@ -0,0 +1,26 @@
+name: Cadence Integration Test
+on:
+  pull_request:
+  push:
+    branches:
+      - 'main'
+
+jobs:
+  tests:
+    name: "Integration testing"
```
longquanzheng, line 10: Let's add `with stickyCache disabled` otherwise it's hard to differentiate with the other one 😂

.github/workflows/ci-temporal-integ-test-disable-sticky.yml
```diff
@@ -0,0 +1,31 @@
+name: Temporal Integration Test
+on:
+  pull_request:
+  push:
+    branches:
+      - 'main'
+
+jobs:
+  tests:
+    name: "Integration testing"
```
longquanzheng, line 10: same here

Makefile
```diff
@@ -177,10 +177,16 @@ cadenceIntegTests:
 	$Q go test -v ./integ -temporal=false
 
 ci-cadence-integ-test:
-	$Q go test -v ./integ -search=false -temporal=false -dependencyWaitSeconds=180
+	$Q go test -v ./integ -search=false -temporal=false -dependencyWaitSeconds=180 -disableStickyCache=true
```
longquanzheng, line 180: Looks like you want to remove `disableStickyCache=true` :P

Makefile
```diff
@@ -177,10 +177,16 @@ cadenceIntegTests:
...
+
+ci-cadence-integ-test-disable-sticky:
```
longquanzheng, line 182: same here

Makefile
```diff
@@ -177,10 +177,16 @@ cadenceIntegTests:
...
+	$Q go test -v ./integ -run "(?i)^Test[${startsWith}]" -search=false -temporal=false -dependencyWaitSeconds=180 -disableStickyCache=true
 
 ci-temporal-integ-test:
```
longquanzheng, line 185: Can you test does `-disableStickyCache=true` work? I thought it has to be `-disableStickyCache` because it's a bool flag

service/interpreter/cadence/worker.go
```diff
@@ -55,6 +63,11 @@ func (iw *InterpreterWorker) Start() {
 		options.MaxConcurrentDecisionTaskPollers = 10
 	}
 
+	// When DisableStickyCache is true it can harm performance; should not be used in production environment
+	if disableStickyCache {
+		options.DisableStickyExecution = true
```
longquanzheng, line 68: Nit: maybe print a log here for confirming the behavior

service/interpreter/temporal/worker.go
```diff
@@ -55,6 +63,11 @@ func (iw *InterpreterWorker) Start() {
 		options.MaxConcurrentWorkflowTaskPollers = 10
 	}
 
+	// When DisableStickyCache is true it can harm performance; should not be used in production environment
+	if disableStickyCache {
+		worker.SetStickyWorkflowCacheSize(0)
```
longquanzheng, line 68: same here :D

Makefile
lwolczynski: It does work, but I can simplify it to `disableStickyCache`

.github/workflows/ci-cadence-integ-test-disable-sticky.yml
lwolczynski: Done

Makefile
lwolczynski: My bad. Copying and pasting things around...

Makefile
lwolczynski: `-disableStickyCache` by itself does not work, so I'm reverting to `-disableStickyCache=true`

lwolczynski: Done

longquanzheng (approved):

PR #476, opened 2024-11-07
IWF-122: adding set search attributes and set search objects apis

ktrops:
### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related Issue
Closes #issue_number


samuel27m:
I took a quick look at the code and I don't think you did anything wrong here.

The error seems to be something with the Docker images:
```
 temporal-admin-tools Error manifest for temporalio/admin-tools:1.25 not found: manifest unknown: manifest unknown
```

Could be that we're due for some updates, unsure. I'll take a quick look into it to see if it's something with an easy fix

samuel27m:
Yes, running locally should reproduce this. However, Docker will cache the downloaded images, and the CI pipelines will always retrieve fresh ones. 

TLDR: to reproduce locally you need to delete the cached image of `temporalio/admin-tools`.

This seems to be the cause of the issue: 
https://github.com/temporalio/temporal/issues/6052#issuecomment-2151164699

Looking at the tags, it is correct that the tag no longer exists: https://hub.docker.com/r/temporalio/admin-tools/tags - they might've purged them recently

lwolczynski:
@samuel27m can you try changing the local variable locally? From `TEMPORAL_VERSION=1.25` to `TEMPORAL_VERSION=1.25.2`

samuel27m:
> @samuel27m can you try changing the local variable locally? From `TEMPORAL_VERSION=1.25` to `TEMPORAL_VERSION=1.25.2`

Doesn't work because the `temporalio/admin-tools` doesn't have that version available. Closest is: `1.25.0-tctl-1.18.1-cli-1.1.0` 😕 

```
manifest for temporalio/admin-tools:1.25 not found: manifest unknown: manifest unknown 
```

samuel27m:
@longquanzheng I'm not sure if we should update the `temporalio/admin-tools` tag to [1.25.0-tctl-1.18.1-cli-1.0.0](https://hub.docker.com/layers/temporalio/admin-tools/1.25.0-tctl-1.18.1-cli-1.0.0/images/sha256-5d3fb34c3db120534c3cd4d0e2f7f2c0d8753aa4ba544a7fea2d54b2b2e8f597?context=explore) or to [1.25.0-tctl-1.18.1-cli-1.1.0](https://hub.docker.com/layers/temporalio/admin-tools/1.25.0-tctl-1.18.1-cli-1.1.0/images/sha256-20ef246194e19f247b82db9217e226104a8a23df149e753e59ee220db7dd8359?context=explore) to fix the tests here. 

Maybe you know which CLI version we should be on? 🤔

longquanzheng (approved):

lwolczynski (approved):

PR #479, opened 2024-11-13
IWF-324: adding keyword array to search attributes

ktrops:
### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related Issue
Closes #issue_number

### DIFF
--- a/docker-compose/init-compose.sh
+++ b/docker-compose/init-compose.sh
@@ -19,7 +19,9 @@ for run in {1..60}; do
   sleep 0.1 
   temporal  operator search-attribute  create --name IwfExecutingStateIds --type KeywordList
   sleep 0.1
-  if checkExists "IwfWorkflowType" ] && checkExists "IwfGlobalWorkflowVersion" && checkExists "IwfExecutingStateIds" ] ; then
+  temporal  operator search-attribute  create --name CustomKeywordArrayField --type KeywordList
+  sleep 0.1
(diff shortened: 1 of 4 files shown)


docker-compose/init-compose.sh
```diff
@@ -19,7 +19,9 @@ for run in {1..60}; do
   sleep 0.1 
   temporal  operator search-attribute  create --name IwfExecutingStateIds --type KeywordList
   sleep 0.1
-  if checkExists "IwfWorkflowType" ] && checkExists "IwfGlobalWorkflowVersion" && checkExists "IwfExecutingStateIds" ] ; then
+  temporal  operator search-attribute  create --name CustomKeywordArrayField --type KeywordList
```
samuel27m, line 22: You don't need to add it here, other than that this looks good 👍

samuel27m, line 22: You don't need to add it here, other than that this looks good 👍

samuel27m (changes requested):

docker-compose/init-compose.sh
ktrops: init-compose.sh is used if we want to setup the docker containers, but run iwf-server locally to test local changes. If some one runs the server locally then runs java-sdk tests against it the tests will fail.

samuel27m: Ah I see, makes sense! Thanks for clarifying, appreciate it! 🙌

samuel27m (approved):

PR #482, opened 2024-11-15
Fix continueAsNew with large snapshot(>4MB)

### LINKED ISSUE #397: Reliability enhancement: move dump internal pagination to query method (IWF-130)
This is to workaround query limit in temporal (4mb) -- today it will fail and keep retrying when needs to continueAsNew with >4MB data as internal states

longquanzheng:
### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related Issue
Closes #397


.github/workflows/ci-cadence-integ-test-disable-sticky.yml
```diff
@@ -13,14 +13,15 @@ jobs:
...
-    steps:
-      - uses: actions/checkout@v3
-      - name: "Set up cadence environment"
-        run: docker compose -f docker-compose/ci-cadence-dependencies.yml up -d
-      - name: "Test against cadence"
-        run: make ci-cadence-integ-test-disable-sticky startsWith=${{ matrix['test-subset'] }}
-      - name: Dump docker logs
-        if: always()
-        uses: jwalton/gh-docker-logs@v2
+# TODO cadence sticky test is flaky
```
longquanzheng, line 18: cc @lwolczynski disabled this as you saw today that it started to be flaky

service/interpreter/persistence.go
```diff
@@ -158,16 +159,22 @@ func (am *PersistenceManager) LoadDataObjects(
 
 func (am *PersistenceManager) GetAllSearchAttributes() []iwfidl.SearchAttribute {
 	var res []iwfidl.SearchAttribute
-	for _, value := range am.searchAttributes {
-		res = append(res, value)
+	// NOTE: using DeterministicKeys so that the JSON snapshot for continueAsNew is stable for pagination
+	// TODO: we should use DeterministicKeys for every map iteration in interpreter for safety
+	// https://github.com/indeedeng/iwf/issues/510
+	for _, k := range DeterministicKeys(am.searchAttributes) {
```
longquanzheng, line 165: This is necessary otherwise the snapshot will have different order for different pages. 
It wasn't a problem becase we didn't see more than one page for continueAsNew.

service/interpreter/continueAsNewer.go
```diff
@@ -110,27 +113,56 @@ func LoadInternalsFromPreviousRun(
...
-			StateExecutionCounterInfo:  c.stateExecutionCounter.Dump(),
-			DataObjects:                c.persistenceManager.GetAllDataObjects(),
-			SearchAttributes:           c.persistenceManager.GetAllSearchAttributes(),
-			StatesToStartFromBeginning: c.stateRequestQueue.GetAllStateStartRequests(),
-			StateExecutionsToResume:    localStateExecutionToResumeMap,
-			StateOutputs:               c.outputCollector.GetAll(),
-			StaleSkipTimerSignals:      c.timerProcessor.Dump(),
-		}, nil
-	})
+	return c.provider.SetQueryHandler(ctx, service.ContinueAsNewDumpQueryType,
```
ktrops, line 138: I'm having a hard time understanding how/where this pagination code (SetQueryHandler) is called when the workflow is queried. What is the code path?

ktrops (approved):

service/interpreter/continueAsNewer.go
longquanzheng: better to use your own IDE to browse the code path:
* workflowImpl.go: `if input.IsResumeFromContinueAsNew { LoadInternalsFromPreviousRun(...) }`
* continueAsNewer.go: `provider.ExecuteActivity(&resp, false, ctx, DumpWorkflowInternal...`
* activityImpl.go: `apiClient.DefaultApi.ApiV1WorkflowInternalDumpPost(ctx)`
* service.go: `s.client.QueryWorkflow(ctx, &pageOfSnapshot, request.GetWorkflowId(), request.GetWorkflowRunId(), service.ContinueAsNewDumpQueryType, request)`
* continueAsNewer.go : the registered handler of `SetQueryHandler(ctx, service.ContinueAsNewDumpQueryType,`

integ/large_data_attributes_test.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func doTestLargeQueryAttributes(t *testing.T, backendType service.BackendType, config *iwfidl.WorkflowConfig) {
+	if !*temporalIntegTest {
+		t.Skip()
+	}
+	assertions := assert.New(t)
+
+	// start test workflow server
+	wfHandler := signal.NewHandler()
```
samuel27m, line 35: Nit: Wouldn't it be better to create a handler class for this specific test?

service/api/handler.go
```diff
@@ -346,8 +346,11 @@ func (h *handler) processError(c *gin.Context, resp *errors.ErrorAndStatus) {
 	c.JSON(resp.StatusCode, resp.Error)
 }
 
-func (h *handler) toJson(req any) string {
+func (h *handler) toJsonLogging(req any) string {
 	str, err := json.Marshal(req)
+	if len(str) > 1000 {
+		str = str[0:1000]
+	}
```
samuel27m, line 353: I thought this logging went to Datadog, is there a specific reason why we need to limit its size? I don't see a reason to limit this if it's not going to Temporal

samuel27m (approved):

service/interpreter/continueAsNewer.go
```diff
@@ -110,27 +113,56 @@ func LoadInternalsFromPreviousRun(
...
+			wholeData, err := json.Marshal(wholeSnapshot)
+			if err != nil {
+				return nil, err
+			}
+			checksum := md5.Sum(wholeData)
+			pageSize := int32(service.DefaultContinueAsNewPageSizeInBytes)
+			if request.PageSizeInBytes > 0 {
+				pageSize = request.PageSizeInBytes
+			}
+			lenInDouble := float64(len(wholeData))
```
lwolczynski, line 150: Nit: no need to declare this since it's only used in line 151. I also believe they could be float32s since you are really dividing ints and just doing type conversion

integ/large_data_attributes_test.go
longquanzheng: I feel like using the existing is better. The test only need a workflow to do nothing, so that we can set some large DAs. The existing signal workflow is exactly like that

service/api/handler.go
longquanzheng: This is for the request object. Logging will be too expensive too if the req is too big (>100 bytes).

This one is actually not going to Datadog because we have another internal handler (see `handler.go` in internal repo). 

I wanted to limit this otherwise my local dev will look pretty bad when sending the 1MB request.

service/interpreter/continueAsNewer.go
longquanzheng: I think having two lines is better for reading. 
The compiler should be smart enough to optimize this to save a local varable. 

float64 is required by the library. Golang doesn't allow int32 to divided by float like Java. 

Screenshot 2024-12-09 at 3 42 27 PM

lwolczynski: Screenshot 2024-12-10 at 9 53 45 AM

Okay, I see that Ceil requires float64, interesting... TIL

PR #485, opened 2024-11-19
IWF-141: Disable global version search attribute

### LINKED ISSUE #427: Disable global version search attribute(IwfGlobalWorkflowVersion) to save an action
Golang SDK has already a built in search attribute “TemporalChangeVersion” to use, and it’s free in the cloud. (note: this is not available in Java SDK https://github.com/temporalio/sdk-java/issues/587#issuecomment-2398907995 ) . Therefore, it is okay to disable our search attribute. 

This needs versioning if doing by default

lwolczynski:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing
- [ ] **This PR CONTAINS a (planned) breaking change** (it is not backward compatible)

### Related Issue
Closes #427

### DIFF
--- a/service/interpreter/globalVersioner.go
+++ b/service/interpreter/globalVersioner.go
@@ -12,7 +12,8 @@ const StartingVersionRenamedStateApi = 3
 const StartingVersionContinueAsNewOnNoStates = 4
 const StartingVersionTemporal26SDK = 5
 const StartingVersionExecutingStateIdMode = 6
-const MaxOfAllVersions = StartingVersionExecutingStateIdMode
+const StartingVersionNoIwfGlobalVersionSearchAttribute = 7
+const MaxOfAllVersions = StartingVersionNoIwfGlobalVersionSearchAttribute
 
 // GlobalVersioner see https://stackoverflow.com/questions/73941723/what-is-a-good-way-pattern-to-use-temporal-cadence-versioning-api
 type GlobalVersioner struct {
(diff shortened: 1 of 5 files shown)


service/interpreter/globalVersioner.go
```diff
@@ -58,6 +59,10 @@ func (p *GlobalVersioner) IsAfterVersionOfTemporal26SDK() bool {
 	return p.version >= StartingVersionTemporal26SDK
 }
 
+func (p *GlobalVersioner) IsAfterVersionOfNoIwfGlobalVersionSearchAttribute() bool {
```
longquanzheng, line 62: actually, we don't have to strictly use this naming pattern. I think we could just rename the existing one to ```go
IsUsingGlobalVersionSearchAttribute(){
   return version >= .. && <= ...; 
}

service/interpreter/globalVersioner.go
```diff
@@ -58,6 +61,16 @@ func (p *GlobalVersioner) IsAfterVersionOfTemporal26SDK() bool {
...
```
longquanzheng, line 64: Nit: it is not being used so can be removed. Do you put it here intentionally for consistency?

longquanzheng (approved):
Lgtm with a nit comment

service/interpreter/globalVersioner.go
lwolczynski: Yes, I added it for consistency

longquanzheng: Yeah Sounds good. I am fine to keep it

PR #491, opened 2024-11-22
IWF-163: Add retry policy to QueryWorkflow calls

### LINKED ISSUE #437: Refactor persistence_test and move API call retry to API Service
Under certain circumstances (starting a continuedAsNew workflow is one of them) it might take time before Temporal/Candace API calls can be resolved successfully and expected data is returned. We should refactor the code and retry API calls when a certain response is received (for example: 500 response)

**ACs:**

- Retry API calls when certain responses are received
  - TODO: Need to decide on checking error types, and backoff strategy
- Refactor [persistence_test.go](https://github.com/indeedeng/iwf/blob/main/integ/persistence_test.go) so it does not need to use Sleep.time method

lwolczynski:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #437

### DIFF
--- a/cmd/server/iwf/iwf.go
+++ b/cmd/server/iwf/iwf.go
@@ -124,11 +124,7 @@ func start(c *cli.Context) {
 		if err != nil {
 			rawLog.Fatalf("Unable to connect to Temporal because of error %v", err)
 		}
-		retryPolicy := temporalapi.QueryWorkflowFailedRetryPolicy{
-			InitialIntervalSeconds: config.GetInitialIntervalSecondsWithDefault(),
-			MaximumAttempts:        config.GetMaximumAttemptsWithDefault(),
-		}
-		unifiedClient = temporalapi.NewTemporalClient(temporalClient, config.Interpreter.Temporal.Namespace, converter.GetDefaultDataConverter(), false, retryPolicy)
+		unifiedClient = temporalapi.NewTemporalClient(temporalClient, config.Interpreter.Temporal.Namespace, converter.GetDefaultDataConverter(), false, &config.Api.QueryWorkflowFailedRetryPolicy)
 
 		for _, svcName := range services {
 			go launchTemporalService(svcName, *config, unifiedClient, temporalClient, logger)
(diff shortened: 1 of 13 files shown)


cmd/server/iwf/iwf.go
```diff
@@ -124,7 +124,11 @@ func start(c *cli.Context) {
 		if err != nil {
 			rawLog.Fatalf("Unable to connect to Temporal because of error %v", err)
 		}
-		unifiedClient = temporalapi.NewTemporalClient(temporalClient, config.Interpreter.Temporal.Namespace, converter.GetDefaultDataConverter(), false)
+		retryPolicy := temporalapi.QueryWorkflowFailedRetryPolicy{
+			InitialIntervalSeconds: config.GetInitialIntervalSecondsWithDefault(),
+			MaximumAttempts:        config.GetMaximumAttemptsWithDefault(),
+		}
+		unifiedClient = temporalapi.NewTemporalClient(temporalClient, config.Interpreter.Temporal.Namespace, converter.GetDefaultDataConverter(), false, retryPolicy)
```
longquanzheng, line 131: If you have `config.GetInitialIntervalSecondsWithDefault()`, then the expectation is that the code that need to use the value will call it to get the value with default, instead of passing into here. 

So it will be `temporalapi.NewTemporalClient(temporalClient, config.Interpreter.Temporal.Namespace, converter.GetDefaultDataConverter(), false, config.ApiConfig.QueryRetryPolicy)`

Also consider using pointers to make this optional and load the default values in `client.go`.
Basically the config can be defined it as a pointer instead of struct. So that internally, we just need to pass a `nil` into this, which should let it use the default value.

service/client/cadence/client.go
```diff
@@ -227,9 +240,24 @@ func (t *cadenceClient) ListWorkflow(
...
-		return err
+	var qres encoded.Value
+	var err error
+
+	attempt := 1
+	for attempt <= t.queryWorkflowFailedRetryPolicy.MaximumAttempts {
+		qres, err = queryWorkflowWithStrongConsistency(t, ctx, workflowID, runID, queryType, args)
+		if err != nil {
+			if t.isQueryFailedError(err) {
+				if attempt == t.queryWorkflowFailedRetryPolicy.MaximumAttempts {
```
longquanzheng, line 251: Nit: I don't think we need another check here, since it'a already checked by the for loop condition?

Although we will have to check `err != nil` again outside of the loop

lwolczynski: Boils down to the same thing, but it might be a little cleaner. Done

service/client/temporal/client.go
```diff
@@ -20,23 +21,45 @@ import (
...
+	queryWorkflowFailedRetryPolicy QueryWorkflowFailedRetryPolicy
 }
 
 func NewTemporalClient(
-	tClient client.Client, namespace string, dataConverter converter.DataConverter, memoEncryption bool,
+	tClient client.Client, namespace string, dataConverter converter.DataConverter, memoEncryption bool, retryPolicy *config.QueryWorkflowFailedRetryPolicy,
 ) uclient.UnifiedClient {
+	var rp QueryWorkflowFailedRetryPolicy
+
+	if retryPolicy.InitialIntervalSeconds == 0 {
```
lwolczynski, line 45: If I'm not mistaken, not providing `Config.ApiConfig.QueryWorkflowFailedRetryPolicy` in yaml file will make the default int value show (which is 0). In other words, we shouldn't need a nil check here

PR #497, opened 2024-11-25
IWF-124: Ignore empty output for collector

longquanzheng:
### Description

Before
Screenshot 2024-11-25 at 3 50 02 PM

After
Screenshot 2024-11-25 at 3 51 09 PM

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #issue_number


ktrops (commented):
If the state doesn't have any output, then it'll be empty? I was using it to know what workflow execution completed. I guess the state id count could work for that.

longquanzheng:
> If the state doesn't have any output, then it'll be empty? I was using it to know what workflow execution completed. I guess the state id count could work for that.

waiting on workflow execution will still work, it doesn't rely on this. 
What "maybe" a small breaking change is that the output of the waitForWorkflowCompletion won't have the stateId/stateExecutionId anymore for those who completing the "threads".

stevo89519:
> What "maybe" a small breaking change is that the output of the waitForWorkflowCompletion won't have the stateId/stateExecutionId anymore for those who completing the "threads".

Is this a concern we need to address?

longquanzheng:
> > What "maybe" a small breaking change is that the output of the waitForWorkflowCompletion won't have the stateId/stateExecutionId anymore for those who completing the "threads".
> 
> Is this a concern we need to address?

I think we can ignore. 
The issue is only in theory but not practical. 
We have this API in java SDK https://github.com/indeedeng/iwf-java-sdk/blob/055f2836ebc375b569b28ea7b45318b847c9ea2b/src/main/java/io/iworkflow/core/Client.java#L427

`getComplexWorkflowResultWithWait` it will return all the completed states even the completion output is empty. So in theory, if someone rely on this in production, it's a breaking changes. But I don't think anyone would do that in practile (I don't see anyone using this API in Indeed, or samples at all, and I doubt any OSS users even aware of this API).

cc @lwolczynski

PR #505, opened 2024-12-03
IWF-372: Fix GetSearchAttributes returning nil when no IwfExecutionStateIds are set

lwolczynski:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #issue_number


service/common/mapper/searchAttribute.go
```diff
@@ -45,7 +45,8 @@ func MapToInternalSearchAttributes(attributes []iwfidl.SearchAttribute) (map[str
 
 func MapCadenceToIwfSearchAttributes(searchAttributes *shared.SearchAttributes, requestedSearchAttributes []iwfidl.SearchAttributeKeyAndType) (map[string]iwfidl.SearchAttribute, error) {
 	if searchAttributes == nil || len(requestedSearchAttributes) == 0 {
-		return nil, nil
+		// return empty map rather than nil
```
longquanzheng, line 48:
```
unable to convert value <nil> to type KEYWORD_ARRAY for key IwfExecutingStateIds
```
The error was when we converting a nil to KeywordArray, so I believe it's at line 66...

You should be able to reproduce this, if you modify `wait_until_search_attributes_optimization_test.go` to make it more complex -- for the default mode (ENABLED_FOR_STATES_WITH_WAIT_UNTIL_ONLY) , after the S7-1, start another S8 with waitUntil and it will try to refresh `IwfExecutingStateIds` with null value...

Screenshot 2024-12-03 at 2 41 02 PM

longquanzheng: Basically, whenever it tried to refresh `IwfExecutingStateIds` that is nil. At the beginning this SA doesn't even exist so there wasn't error. 

Then it was set to some values(array), and then cleared up as a nil. Then setting again will see the errors.

longquanzheng:
I also think we should change `refreshIwfExecutingStateIdSearchAttribute` to return the error instead of logging it. 

Potentially, swallowing error and proceeding will cause wrong behavior -- we were super lucky last time that swallowing error was correct in that case 😂 

By returning error, the workflow will be blocked and we have to fix the bug.

service/common/mapper/searchAttribute.go
longquanzheng: An easy way to produce this:
S1->S2->S1 where S1 has waitUntil but S2 doesn't have.

lwolczynski: Fixed + made an existing test to verify it works as expected

longquanzheng (approved):

PR #456, opened 2024-10-17
Add documetaion to integ workflow implemented by raw REST controller

longquanzheng:
### Checklist
- [ ] Code compiles correctly
- [ ] Tests for the changes have been added
- [ ] All tests passing
- [ ] **This PR change is backwards-compatible**
- [ ] **This PR CONTAINS a (planned) breaking change** (it is not backwards compatible)

### Related Issue
Closes #issue_number

PR #515, opened 2024-12-10
iwf-401: update activity type name to find reset history eventID

ktrops:
### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related Issue
Closes #issue_number

### DIFF
--- a/integ/workflow/reset/routers.go
+++ b/integ/workflow/reset/routers.go
@@ -1,0 +1,121 @@
+package reset
+
+import (
+	"fmt"
+	"github.com/gin-gonic/gin"
+	"github.com/indeedeng/iwf/gen/iwfidl"
+	"github.com/indeedeng/iwf/service"
+	"log"
+	"net/http"
+	"strconv"
+)
+
+/**
+* This test workflow has 2 states, using REST controller to implement the workflow directly.
+* State1:
+*       - No WaitUntil
+*       - Execute moves to State2
+* State2:
+* 		- No WaitUntil
+*       - Execute loops through state2 5 times, then gracefully completes the workflow.
+* This test is used for testing reset by state id and state execution id without WaitUntil
+ */
+const (
+	WorkflowType = "reset"
+	State1       = "S1"
+	State2       = "S2"
+)
+
+type handler struct {
+	invokeHistory map[string]int64
+}
+
+func NewHandler() *handler {
+	return &handler{
+		invokeHistory: make(map[string]int64),
+	}
+}
+
+// ApiV1WorkflowStartPost - for a workflow
+func (h *handler) ApiV1WorkflowStateStart(c *gin.Context) {
+	panic("No start call is expected.")
+}
+
+func (h *handler) ApiV1WorkflowStateDecide(c *gin.Context) {
+	log.Println("start of decide")
+	var req iwfidl.WorkflowStateDecideRequest
+	if err := c.ShouldBindJSON(&req); err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		return
+	}
+	log.Println("received state decide request, ", req)
+	context := req.GetContext()
+	if context.GetAttempt() <= 0 || context.GetFirstAttemptTimestamp() <= 0 {
+		panic("attempt and firstAttemptTimestamp should be greater than zero")
+	}
+
+	if req.GetWorkflowType() == WorkflowType {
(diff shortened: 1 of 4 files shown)


longquanzheng:
PR title: "iwf-401: update activity type name" -> "iwf-401: update activity type name to find reset history eventID"

This title will become the commit message so it's better to make it as clear as possible

longquanzheng:
Also need to do it for Cadence

longquanzheng:
Unless its too hard, I would recommend you to add a test. 

I have added one here in the signal test: https://github.com/indeedeng/iwf/blob/main/integ/signal_test.go#L236

it reset to stateId/stateExecutionId but the state has `waitUntil` so it doesn't cover the case that you fixed. So you can just find a workflow state that doesn't have waitUntil and reset it, then wait for the workflow to complete.

stevo89519:
logic looks good

longquanzheng:
https://github.com/indeedeng/iwf/blob/main/integ/workflow/skipstart/routers.go

this is a test workflow that skip waitUntil(used to be called stateStart)

integ/workflow/reset/routers.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/indeedeng/iwf/gen/iwfidl"
+	"github.com/indeedeng/iwf/service"
+	"log"
+	"net/http"
+	"strconv"
+)
+
+//A workflow without calling WaitUntil api for any state. It loops through state2 5 times. Used for:
+// * Testing reset by state id and state execution id.
```
longquanzheng, line 14: I still need to document all the integ test like this format from @samuel27m https://github.com/indeedeng/iwf/pull/456/files

It would be nice to follow so that it's much more clear to know the workflow

integ/workflow/reset/routers.go
```diff
@@ -0,0 +1,168 @@
...
+		if req.GetWorkflowStateId() == State1 || req.GetWorkflowStateId() == State2 {
+			h.invokeHistory[req.GetWorkflowStateId()+"_start"]++
+			log.Println("beg start response")
+			c.JSON(http.StatusOK, iwfidl.WorkflowStateStartResponse{
+				CommandRequest: &iwfidl.CommandRequest{
+					DeciderTriggerType: iwfidl.ALL_COMMAND_COMPLETED.Ptr(),
+				},
+			})
+			return
+		}
```
longquanzheng, line 57: we don't expect any waitUntil, right? So we can just panic in this method?

integ/workflow/reset/routers.go
```diff
@@ -0,0 +1,168 @@
...
+									BackoffCoefficient:     iwfidl.PtrFloat32(14),
+									MaximumAttempts:        iwfidl.PtrInt32(14),
+									MaximumIntervalSeconds: iwfidl.PtrInt32(14),
+								},
+								ExecuteApiRetryPolicy: &iwfidl.RetryPolicy{
+									InitialIntervalSeconds: iwfidl.PtrInt32(15),
+									BackoffCoefficient:     iwfidl.PtrFloat32(15),
+									MaximumAttempts:        iwfidl.PtrInt32(15),
+									MaximumIntervalSeconds: iwfidl.PtrInt32(15),
+								},
```
longquanzheng, line 102: you only need `SkipWaitUntil: iwfidl.PtrBool(true),` for this test

integ/workflow/reset/routers.go
```diff
@@ -0,0 +1,168 @@
...
+										InitialIntervalSeconds: iwfidl.PtrInt32(14),
+										BackoffCoefficient:     iwfidl.PtrFloat32(14),
+										MaximumAttempts:        iwfidl.PtrInt32(14),
+										MaximumIntervalSeconds: iwfidl.PtrInt32(14),
+									},
+									ExecuteApiRetryPolicy: &iwfidl.RetryPolicy{
+										InitialIntervalSeconds: iwfidl.PtrInt32(15),
+										BackoffCoefficient:     iwfidl.PtrFloat32(15),
+										MaximumAttempts:        iwfidl.PtrInt32(15),
+										MaximumIntervalSeconds: iwfidl.PtrInt32(15),
```
longquanzheng, line 138: same here

iwf-idl
longquanzheng, line 1: I don't think you need to update any IDL. you are probably rolling it back to older version...

longquanzheng (approved):

PR #524, opened 2024-12-16
IWF-124: Ignore empty output for collector

lwolczynski:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #issue_number

### DIFF
--- a/service/interpreter/outputCollector.go
+++ b/service/interpreter/outputCollector.go
@@ -7,22 +7,11 @@ type OutputCollector struct {
 }
 
 func NewOutputCollector(initOutputs []iwfidl.StateCompletionOutput) *OutputCollector {
-	filteredOutputs := []iwfidl.StateCompletionOutput{}
-
 	if initOutputs == nil {
-		return &OutputCollector{
-			outputs: filteredOutputs,
-		}
-	} else {
-		for _, output := range initOutputs {
-			if output.CompletedStateOutput != nil {
-				filteredOutputs = append(filteredOutputs, output)
-			}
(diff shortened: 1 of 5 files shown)


service/interpreter/outputCollector.go
```diff
@@ -7,16 +7,29 @@ type OutputCollector struct {
...
+			if output.CompletedStateOutput != nil {
+				filteredOutputs = append(filteredOutputs, output)
+			}
+		}
+
+		return &OutputCollector{
+			outputs: filteredOutputs,
+		}
 	}
 }
```
lwolczynski, line 27: @longquanzheng is there a reason why we wouldn't filter out outputs with nil CompletedStateOutput here?

lwolczynski: Discussed

longquanzheng (approved):
lgtm, 
do we know we have existing test with output(we probably do)
if no, can we change one of the test to return a non-nil output so that we can test the case?

lwolczynski:
> lgtm, do we know we have existing test with output(we probably do) if no, can we change one of the test to return a non-nil output so that we can test the case?

basic_test.go:
`	assertions.Equal(1, len(resp2.GetResults()))`

conditional_close_test.go:
`	assertions.Equal(1, len(resp2.GetResults()))`
`assertions.Equal(expectedOutput, resp2.GetResults()[0])`

skip_start_test.go:
`	assertions.Equal(1, len(resp2.GetResults()))`

PR #529, opened 2024-12-23
IWF-274: Optimize Timer creation

### LINKED ISSUE #465: Optimize timer started for saving cost
It's a common pattern using iwf workflow like this:
```
class UpdatableTimerState implements WorkflowState<Void> {

    @Override
    public CommandRequest waitUntil(Context context, ...) {
        return CommandRequest.forAnyCommandCompleted(
                TimerCommand.createByDuration(TIMEOUT_DURATION)),
                InternalChannelCommand.create(CHANNEL_RESET_TIMER)
        );
    }

    @Override
    public StateDecision execute(Context context, Void input, CommandResults commandResults,...) {
        TimerCommandResult timer = commandResults.getTimerResult(0);
        if (timer.getStatus() == FIRED) {
              // timer fires, call API to send an email
              svc.sendEmail(...);
              // assuming we want to complete workflow here
              return StateDecision.completeWorkflow();
        }
        // otherwise the channel receives a message to reset the timer
        return StateDecision.singleNextState(UpdatableTimerState.class);
    }
```

When doing this, a new timer would always created when receiving a REST TIMER message. 
We could potentially lazily create it when needed. For example -- 
 if the first timer is 30 days, and after one day, it tries to create another 30 day timer, after another day, it will create another 30 day timer again, so there are 3 timers created in history. 

Instead, we could keep the first timer, and only create a new timer when the first timer is fired, and check if there is another timer needed. 

Note that this optimization is now possible to implement relatively easy because iwf has a dedicated component to manage all the timers -- [TimerProcessor](https://github.com/indeedeng/iwf/blob/main/service/interpreter/timerProcessor.go):
* It contains all the real timers in member fields `stateExecutionCurrentTimerInfos`
  * continueAsNew is implemented already for it. 
* The temporal timers are actually started in a centralized place in `WaitForTimerFiredOrSkipped` method. 

So technically this optimization can be done within this component, without changing anything else. 

It would be safer to enable by config, then later on when it's stable, use versioning to enable for all.

bowersj27:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #465

### DIFF
--- a/integ/greedy_timer_test.go
+++ b/integ/greedy_timer_test.go
@@ -106,13 +106,13 @@ func doTestGreedyTimerWorkflowCustomConfig(t *testing.T, backendType service.Bac
 	time.Sleep(time.Second * 1)
 
 	// assertions
-	timers := service.GetScheduledGreedyTimerTimesQueryResponse{}
-	err = uClient.QueryWorkflow(context.Background(), &timers, wfId, "", service.GetScheduledGreedyTimerTimesQueryType)
+	debug := service.DebugDumpResponse{}
+	err = uClient.QueryWorkflow(context.Background(), &debug, wfId, "", service.DebugDumpQueryType)
 	if err != nil {
 		log.Fatalf("Fail to invoke query %v", err)
 	}
-	assertions.Equal(1, len(timers.ScheduledGreedyTimerTimes))
-	singleTimerScheduled := timers.ScheduledGreedyTimerTimes[0]
+	assertions.Equal(1, len(debug.FiringTimersUnixTimestamps))
+	singleTimerScheduled := debug.FiringTimersUnixTimestamps[0]
 
 	scheduleTimerAndAssertExpectedScheduled(t, apiClient, uClient, wfId, 20, 1)
 
@@ -127,15 +127,15 @@ func doTestGreedyTimerWorkflowCustomConfig(t *testing.T, backendType service.Bac
 
 	time.Sleep(time.Second * 1)
 
-	err = uClient.QueryWorkflow(context.Background(), &timers, wfId, "", service.GetScheduledGreedyTimerTimesQueryType)
+	err = uClient.QueryWorkflow(context.Background(), &debug, wfId, "", service.DebugDumpQueryType)
 	if err != nil {
 		log.Fatalf("Fail to invoke query %v", err)
 	}
 
 	// no second timer started
-	assertions.Equal(1, len(timers.ScheduledGreedyTimerTimes))
+	assertions.Equal(1, len(debug.FiringTimersUnixTimestamps))
 	// LessOrEqual due to continue as new workflow scheduling the next, not skipped timer
-	assertions.LessOrEqual(singleTimerScheduled, timers.ScheduledGreedyTimerTimes[0])
+	assertions.LessOrEqual(singleTimerScheduled, debug.FiringTimersUnixTimestamps[0])
 	scheduleTimerAndAssertExpectedScheduled(t, apiClient, uClient, wfId, 5, 2)
 
 	// wait for the workflow
@@ -178,11 +178,11 @@ func scheduleTimerAndAssertExpectedScheduled(
 
 	time.Sleep(time.Second * 1)
 
-	timers := service.GetScheduledGreedyTimerTimesQueryResponse{}
-	err = uClient.QueryWorkflow(context.Background(), &timers, wfId, "", service.GetScheduledGreedyTimerTimesQueryType)
+	debug := service.DebugDumpResponse{}
+	err = uClient.QueryWorkflow(context.Background(), &debug, wfId, "", service.DebugDumpQueryType)
 	if err != nil {
 		log.Fatalf("Fail to invoke query %v", err)
 	}
 
-	assertions.LessOrEqual(len(timers.ScheduledGreedyTimerTimes), noMoreThan)
+	assertions.LessOrEqual(len(debug.FiringTimersUnixTimestamps), noMoreThan)
 }
--- a/integ/greedy_timer_test.go
+++ b/integ/greedy_timer_test.go
@@ -51,9 +51,8 @@ func TestGreedyTimerWorkflowBaseCadenceContinueAsNew(t *testing.T) {
 		t.Skip()
 	}
 	for i := 0; i < *repeatIntegTest; i++ {
-		// TODO: provide durations
 		doTestGreedyTimerWorkflowCustomConfig(t, service.BackendTypeCadence, greedyTimerConfig(true))
-		// TODO: test stuff
+		smallWaitForFastTest()
 	}
 }
 
--- a/integ/greedy_timer_test.go
+++ b/integ/greedy_timer_test.go
@@ -65,7 +65,7 @@ func doTestGreedyTimerWorkflowCustomConfig(t *testing.T, backendType service.Bac
 	assertions := assert.New(t)
 	// start test workflow server
 	wfHandler := greedy_timer.NewHandler()
-	closeFunc1 := startWorkflowWorkerWithRpc(wfHandler)
+	closeFunc1 := startWorkflowWorkerWithRpc(wfHandler, t)
 	defer closeFunc1()
 
 	uClient, closeFunc2 := startIwfServiceWithClient(backendType)
@@ -102,7 +102,7 @@ func doTestGreedyTimerWorkflowCustomConfig(t *testing.T, backendType service.Bac
 			WorkflowConfigOverride: config,
 		},
 	}).Execute()
-	panicAtHttpError(err, httpResp)
+	failTestAtHttpError(err, httpResp, t)
 
 	time.Sleep(time.Second * 1)
 
@@ -115,7 +115,7 @@ func doTestGreedyTimerWorkflowCustomConfig(t *testing.T, backendType service.Bac
 	assertions.Equal(1, len(timers.ScheduledGreedyTimerTimes))
 	singleTimerScheduled := timers.ScheduledGreedyTimerTimes[0]
 
-	scheduleTimerAndAssertExpectedScheduled(assertions, apiClient, uClient, wfId, 20, 1)
+	scheduleTimerAndAssertExpectedScheduled(t, apiClient, uClient, wfId, 20, 1)
 
 	// skip next timer for state: schedule-1
 	skipReq := apiClient.DefaultApi.ApiV1WorkflowTimerSkipPost(context.Background())
@@ -124,7 +124,7 @@ func doTestGreedyTimerWorkflowCustomConfig(t *testing.T, backendType service.Bac
 		WorkflowStateExecutionId: greedy_timer.ScheduleTimerState + "-1",
 		TimerCommandId:           iwfidl.PtrString("duration-15"),
 	}).Execute()
-	panicAtHttpError(err, httpResp)
+	failTestAtHttpError(err, httpResp, t)
 
 	time.Sleep(time.Second * 1)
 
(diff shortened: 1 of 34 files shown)


service/interpreter/timers/greedyTimerProcessor.go
```diff
@@ -0,0 +1,264 @@
...
+func (t *sortedTimers) removeTimer(toRemove *service.TimerInfo) {
+	for i, timer := range t.timers {
+		if toRemove == timer {
+			t.timers = append(t.timers[:i], t.timers[i+1:]...)
+			return
+		}
+	}
+}
+
+func (t *sortedTimers) pruneToNextTimer(pruneTo int64) *service.TimerInfo {
```
longquanzheng, line 87: looks like it's `prune the pendingSchedules to target time, and also return the earliest time`?
it would be nice add some comment here

service/interpreter/timers/greedyTimerProcessor.go
```diff
@@ -0,0 +1,264 @@
...
+
+	for i := len(t.timers) - 1; i >= 0; i-- {
+		timer := t.timers[i]
+		if timer.FiringUnixTimestampSeconds > pruneTo && timer.Status == t.status {
+			break
+		}
+		index = i
+	}
+	t.timers = t.timers[:index]
+	return t.timers[index-1]
```
longquanzheng, line 103: is it possible to have index = 0 and this will panic?

longquanzheng: looking at the below code, looks like we should return nil here and the below code will handle it

service/interpreter/timers/greedyTimerProcessor.go
```diff
@@ -0,0 +1,264 @@
...
+		for {
+			t.provider.Await(ctx, func() bool {
+				// remove fired timers
+				now := t.provider.Now(ctx).Unix()
+				for i := len(createdTimers) - 1; i >= 0; i-- {
+					if createdTimers[i] > now {
+						createdTimers = createdTimers[:i+1]
+						break
+					}
+				}
```
longquanzheng, line 122: Nit: maybe also move to `pruneScheduledTimes` and rename the above to `prunePendingSchedules`

service/interpreter/timers/greedyTimerProcessor.go
```diff
@@ -0,0 +1,264 @@
...
+				next := t.pendingTimers.pruneToNextTimer(now)
+				return (next != nil && (len(createdTimers) == 0 || next.FiringUnixTimestampSeconds < createdTimers[len(createdTimers)-1])) || continueAsNewCounter.IsThresholdMet()
```
longquanzheng, line 124: maybe add a few comments here for the cases:
* "next != nil && (len(createdTimers) == 0" means we should schedule a next timer because there are none
* "next.FiringUnixTimestampSeconds < createdTimers[len(createdTimers)-1]))“ means we should schedule a next timer because the next pending one is earlier than the earliest in scheduledTimes
* continueAsNewCounter.IsThresholdMet() continueAsNew is needed

service/interpreter/timers/greedyTimerProcessor.go
```diff
@@ -0,0 +1,264 @@
...
+
+			if continueAsNewCounter.IsThresholdMet() {
+				break
+			}
+
+			now := t.provider.Now(ctx).Unix()
+			next := t.pendingTimers.pruneToNextTimer(now)
+			//next := t.pendingTimers.getEarliestTimer()
+			// only create a new timer when a pending timer exists before the next existing timer fires
+			if next != nil && (len(createdTimers) == 0 || next.FiringUnixTimestampSeconds < createdTimers[len(createdTimers)-1]) {
```
longquanzheng, line 135: I feel like we don't have to do this check again, if it's not continueAsNew, it should be the case already?

service/interpreter/timers/greedyTimerProcessor.go
```diff
@@ -0,0 +1,264 @@
...
+				fireAt := next.FiringUnixTimestampSeconds
+				duration := time.Duration(fireAt-now) * time.Second
+				t.provider.NewTimer(ctx, duration)
```
longquanzheng, line 138: This NewTimer is a non-blocking one. 
At first glance I thought it should be `t.provider.Sleep(...)` then I remember that we had a similar discussion when implemeting it for Wayfinder LoL.

So we don't need to wait here because we rely on the fact that when the timer fired, workflow task will get triggered and evaluate the Await condition, right? 

Maybe add more comments here too.

service/interpreter/timers/greedyTimerProcessor.go
```diff
@@ -0,0 +1,264 @@
...
+		return timer.Status
+	}
+	skippedByStaleSkip := t.RetryStaleSkipTimer()
+	if skippedByStaleSkip {
+		t.logger.Warn("timer skipped by stale skip signal", stateExeId, timerIdx)
+		timer.Status = service.TimerSkipped
+		return service.TimerSkipped
+	}
+
+	_ = t.provider.Await(ctx, func() bool {
```
longquanzheng, line 211: Nit: add a comment here that the timers in the TimerScheduler will trigger a workflow task when the timer fire, hence it will drive the workflow task to evalaute `timer.FiringUnixTimestampSeconds <= t.provider.Now(ctx).Unix()`. 

That was a tricky part too in our Wayfinder code :P

longquanzheng (commented):
Very great job!! this is quite complicated, but I believe it's worth it.

longquanzheng:
Would be great to add a new test case to it -- especially the idleTimeout pattern, which is that main use case that this optimization is for. 

Most of the code can be from `any_timer_signal_test`
https://github.com/indeedeng/iwf/blob/main/integ/workflow/any_timer_signal/routers.go#L47
Currently it’s
`S1-1->S1-2->S2-1`

S1-1 will wait for timer, S1-2 will wait for timer+signal. 

We can change it to 
S1-1->S1-2->(maybe a few more iterations) -> completeWF

Both S1-1 and S1-2 will wait for signal+timer and only the second timer will fired. We will expect only one timer was started in Temporal. 
Maybe we have to check history events like this: https://github.com/indeedeng/iwf/blob/main/integ/wait_until_search_attributes_optimization_test.go#L133
(and only doing it for Temporal is fine)

service/interpreter/timers/greedyTimerProcessor.go
```diff
@@ -0,0 +1,264 @@
...
+		insertIndex = i + 1
+	}
+	t.timers = append(
+		t.timers[:insertIndex],
+		append([]*service.TimerInfo{toAdd}, t.timers[insertIndex:]...)...)
+}
+
+func (t *sortedTimers) removeTimer(toRemove *service.TimerInfo) {
+	for i, timer := range t.timers {
+		if toRemove == timer {
```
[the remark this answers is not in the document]
bowersj27: Sorry for getting to this with a delay. I was assuming the reference would be passed through but I do think comparing by value is a safer approach here. I'll make this change.

service/interpreter/timers/greedyTimerProcessor.go
```diff
@@ -0,0 +1,273 @@
...
+
+type TimerManager struct {
+	// Does not map to the timers actually created by the workflow provider
+	PendingScheduling []*service.TimerInfo
+	// timers created through the workflow provider that are going to fire
+	ScheduledTimerTimes []int64
+}
+
+type GreedyTimerProcessor struct {
+	timerManger                     TimerManager
```
lwolczynski, line 20: Typo: `timerManager`

service/interpreter/timers/greedyTimerProcessor.go
```diff
@@ -0,0 +1,264 @@
...
+		pendingTimers:                   sortedTimers{status: service.TimerPending},
+		stateExecutionCurrentTimerInfos: map[string][]*service.TimerInfo{},
+		logger:                          provider.GetLogger(ctx),
+		staleSkipTimerSignals:           staleSkipTimerSignals,
+	}
+
+	// start some single thread that manages timers
+	tp.createGreedyTimerScheduler(ctx, continueAsNewCounter)
+
+	err := provider.SetQueryHandler(ctx, service.GetCurrentTimerInfosQueryType, func() (service.GetCurrentTimerInfosQueryResponse, error) {
```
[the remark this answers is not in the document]
bowersj27: Since all query methods, besides the CaN dump, seem to be for debugging purposes, I added another query method that's only available if the timer optimization is enabled. I can leave the option open to consolidate the two query methods into one in a future iteration if that is cleaner.

service/interpreter/timers/greedyTimerProcessor.go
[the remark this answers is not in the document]
bowersj27: I tested this just in case. This reference compare works here since the timer info in both the `TimerManager` and `stateExecutionCurrentTimerInfos` are storing the same references.

service/interpreter/timers/greedyTimerProcessor.go
bowersj27: I needed to prune both scheduled times and the pending timers in a new func. I named it `pruneToNextTimer`

service/interpreter/timers/greedyTimerProcessor.go
```diff
@@ -0,0 +1,264 @@
...
+			// don't want dupes. Makes remove simpler
+			if toAdd == timer {
+				return
+			}
+			insertIndex = i
+			break
+		}
+		insertIndex = i + 1
+	}
+	t.timers = append(
```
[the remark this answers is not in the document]
bowersj27: the following test worked for me. 
```go
func main() {
    toInsert := 4
    array := []int{1, 2, 3}

    result := append(array[:3], append([]int{toInsert}, array[3:]...)...)

    fmt.Println(result) // Output: [1 2 3 4]
}

```
however your example does result in the `slice bounds out of range` as expected
```go
func main() {
	arr := []int{1, 2}
	fmt.Println(arr)      // Output: [1 2]
	fmt.Println(arr[:1])  // Output: [1]
	fmt.Println(arr[:2])  // Output: [1 2]
	fmt.Println(arr[3:])  // panic: runtime error: slice bounds out of range
}
```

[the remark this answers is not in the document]
bowersj27: I consulted the expert chatgpt on the subject matter: 
```
Slicing with arr[3:]:
- When you use arr[3:] directly, you are attempting to create a slice starting from index 3 to the end of the array or slice.
- If the slice arr has fewer than 3 elements, this will result in a runtime panic due to "slice bounds out of range" because index 3 does not exist.
Using arr[3:]... in append:
- The ... syntax is known as the "variadic argument" syntax in Go. It is used to pass a slice as a series of arguments to a variadic function, such as append.
- When you use arr[3:]... in the context of append, it effectively means "append the elements from index 3 to the end of the slice, if they exist."
- If arr has fewer than 3 elements, arr[3:] results in an empty slice, and append can handle an empty slice without causing a panic.
```
it seems to be true from my testing, but I cannot find this explicitly documented in the docs. This is a strange behavior. I didn't catch/think of this edge case while I implemented this, but out of dumb luck it works due to some reason I still cannot understand. I think having the extra checks in place is desirable to make this read better even if it's equally as correct.

bowersj27:
Moved interfaces to it's own package. There's a fair amount of IDE refactored code because of this. The main changes to review are 
- `workflowImpl`
- `greedyTimerScheduler`
- `greedyTimerProcessor`
- new integTests

service/interpreter/timers/greedyTimerProcessor.go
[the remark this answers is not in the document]
longquanzheng: I like having another query handler. 👍 

But actually all the query types are being used by API except for `DebugDumpQueryType`. 
 
`GetCurrentTimerInfosQueryType` is being used by `skipTimer` API. 
`GetDataAttributesWorkflowQueryType` is being used by getDataAttribute API. `PrepareRpcQueryType` is being used by rpc API, etc. 

Because they are used by API, putting extra debugging data will not work well (wasting data transportation, and could exceed grpc limit). So I agreed not putting the info into this GetCurrentTimerInfosQueryType.

Also, I was thinking about having a dedicated dump query handler. It's called `DebugDumpQueryType` like what you did in the wayfinder one. That's the only one that being used for debugging only today and not being used by service API.

integ/greedy_timer_test.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func TestGreedyTimerWorkflowBaseCadenceContinueAsNew(t *testing.T) {
+	if !*cadenceIntegTest {
+		t.Skip()
+	}
+	for i := 0; i < *repeatIntegTest; i++ {
+		// TODO: provide durations
+		doTestGreedyTimerWorkflowCustomConfig(t, service.BackendTypeCadence, greedyTimerConfig(true))
+		// TODO: test stuff
```
lwolczynski, line 56: 👀

lwolczynski: Did you want to tackle these?

bowersj27: woops, no. That was a prior idea for a test that I didn't go with

lwolczynski (approved):

PR #540, opened 2025-02-05
IWF-532: Close integration test workflows when finished

### LINKED ISSUE #537: Each Integration Test should close their workflows when finished
As part of our Integration Test suite, there are a lot of tests, and all of them create 1 workflow or more, runs assertions, and marks the test as passed. 

This has lead to [situations where workflows are left open](https://github.com/indeedeng/iwf/pull/535#discussion_r1932832641), and then they are trying to interact with another tests worker, which fails the tests. E.g.:

`Test1` starts the worker and service
Kicks off 2 workflows
2nd doesn't start, cool, move on, that's what we wanted to test
close the worker and service
workflow might still be running, waiting for workers

`Test2` starts the worker and service
Runs its workflow
Waits for a while (for the timeout to happen) in`Test2` workflow
The running workflow from `Test1` finds a worker running, and kicks off requests
test fails because it didn't expect the call from the other workflow

The solution here is to ensure that each Integration Test closes the workflows it created before finishing.

lwolczynski:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #537

### DIFF
--- a/integ/persistence_loading_policy_test.go
+++ b/integ/persistence_loading_policy_test.go
@@ -158,6 +158,13 @@ func doTestPersistenceLoadingPolicy(
 		"S2_start":  1,
 		"S2_decide": 1,
 	}, history, "persistence loading policy test fail, %v", history)
+
+	// Terminate the workflow once tests completed
+	stopReq := apiClient.DefaultApi.ApiV1WorkflowStopPost(context.Background())
+	_, err = stopReq.WorkflowStopRequest(iwfidl.WorkflowStopRequest{
+		WorkflowId: wfId,
+		StopType:   iwfidl.TERMINATE.Ptr(),
+	}).Execute()
 }
 
 func getSearchAttributesToGetFromMemo(loadingType iwfidl.PersistenceLoadingType) []iwfidl.SearchAttributeKeyAndType {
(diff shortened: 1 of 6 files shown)


bowersj27 (approved):
LGTM. I think we should consider how we might effectively setup and teardown these integ tests' dependent environments in `integ/util.go:dostartIwfServiceWithClient` to achieve this in the future.   
  
`integ/util.go:dostartIwfServiceWithClient` AFAICT sets up the iwf process and connection to the backing provider(cadence/temporal) for each test. We may be able to expand on this to track all started wf's and make sure they are closed on each test execution such that we don't have to explicitly add this behavior to each test in the future.

samuel27m (approved):

lwolczynski:
> LGTM. I think we should consider how we might effectively setup and teardown these integ tests' dependent environments in `integ/util.go:dostartIwfServiceWithClient` to achieve this in the future.
> 
> `integ/util.go:dostartIwfServiceWithClient` AFAICT sets up the iwf process and connection to the backing provider(cadence/temporal) for each test. We may be able to expand on this to track all started wf's and make sure they are closed on each test execution such that we don't have to explicitly add this behavior to each test in the future.

Opened #542 for it

PR #543, opened 2025-02-10
IWF-511: Add a error field to the IwfEvent struct

samuel27m:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #issue_number


service/client/interfaces.go
```diff
@@ -41,6 +41,7 @@ type UnifiedClient interface {
 type errorHandler interface {
 	GetApplicationErrorTypeIfIsApplicationError(err error) string
 	GetApplicationErrorDetails(err error, detailsPtr interface{}) error
+	GetApplicationErrorTypeAndDetails(err error) (string, string)
```
samuel27m, line 44: Looking for feedback here. Is this the best place for this to live in?
I need to call `GetApplicationErrorTypeIfIsApplicationError` and `GetApplicationErrorDetails` to get the error details for the logging, and I thought adding a function here that combines the two would be the best way. 

But maybe it isn't, because the implementation between temporal and cadence are the exact same, perhaps it belongs in an utils file or something similar. I'm unsure here, please let me know what you think!

samuel27m:
Closing this PR. After looking closer at the test failures, this impl doesn't work for all failure types

samuel27m:
Closing this PR. After looking closer at the test failures, this impl doesn't work for all failure types
Will open a new one when I figure out how to fix that

PR #545, opened 2025-02-12
IWF-343: Fix concurrent write panic for parallel test

### LINKED ISSUE #472: Fix concurrent write panic for parallel test
Maybe using sync.Map: https://medium.com/@luanrubensf/concurrent-map-access-in-go-a6a733c5ffd1

samuel27m:
### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related Issue
Closes #472


integ/workflow/parallel/routers.go
```diff
@@ -71,7 +72,11 @@ func (h *handler) ApiV1WorkflowStateStart(c *gin.Context, t *testing.T) {
 	log.Println("received state start request, ", req)
 
 	if req.GetWorkflowType() == WorkflowType {
-		h.invokeHistory[req.GetWorkflowStateId()+"_start"]++
+		if value, ok := h.invokeHistory.Load(req.GetWorkflowStateId() + "_start"); ok {
+			h.invokeHistory.Store(req.GetWorkflowStateId()+"_start", value.(int64)+1)
+		} else {
+			h.invokeHistory.Store(req.GetWorkflowStateId()+"_start", int64(1))
+		}
```
samuel27m, line 79: I couldn't reproduce the issue described, but I can see that this issue could very well happen due to the nature of the Workflow in this test.

Looking for opinions on if this is something that we need to polish a bit more, and make all our other tests also use `sync.Map` for storing the current `invokeHistory` in the integration tests

samuel27m: As discussed, I have applied the same logic now for all the tests. i.e. all tests now use `sync.Map` which can be written to concurrently

integ/any_command_combination_test.go
```diff
@@ -10,7 +10,7 @@ import (
 	"time"
 
 	"github.com/indeedeng/iwf/gen/iwfidl"
-	anycommandconbination "github.com/indeedeng/iwf/integ/workflow/any_command_combination"
+	anycommandcombination "github.com/indeedeng/iwf/integ/workflow/any_command_combination"
```
samuel27m, line 13: This was just a typo that I noticed while working on the tests

PR #548, opened 2025-02-21
IWF-473: Upgrade Cadence dependencies

### LINKED ISSUE #432: Upgrade cadence dependencies
https://github.com/uber/cadence/blob/master/docker/docker-compose-es-v7.yml

Cadence has now released to 1.x but we are using 0.24.0 

We should upgrade it and also upgrade the Kafka images accordingly

samuel27m:
### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related Issue
Closes #432


integ/workflow/wf_ignore_already_started/routers.go
```diff
@@ -55,12 +50,6 @@ func (h *handler) ApiV1WorkflowStateStart(c *gin.Context, t *testing.T) {
 		}
 
 		if req.GetWorkflowStateId() == State1 {
-			nowInt, err := strconv.Atoi(req.StateInput.GetData())
```
samuel27m, line 58: This was not being used in the test assertions and it was causing this test to be flaky for some weird reason, so I removed it

docker-compose/ci-cadence-dependencies.yml
```diff
@@ -18,43 +18,45 @@ services:
...
     ports:
       - "9042:9042"
+    healthcheck:
+      test: [ "CMD", "cqlsh", "-u cassandra", "-p cassandra", "-e describe keyspaces" ]
+      interval: 15s
+      timeout: 30s
+      retries: 10
     networks:
       - testing-network
-  zookeeper:
```
mixydubs, line 26: Looks like zookeeper is no longer a dependency with the latest version of Cadence, is that right?

integ/workflow/wf_ignore_already_started/routers.go
mixydubs: i think that's for the best, if the value isn't being used but is causing issues it should be removed 👍

mixydubs (approved):
Looks ok from my perspective.  The tests all pass and you've improved the code in few areas while your at it, nice one!

docker-compose/ci-cadence-dependencies.yml
samuel27m: Yes, this is the cadence MR that removed Zookeeper for reference: https://github.com/cadence-workflow/cadence/commit/4cb4443f6647c730430e1131780e6674f5eed3cf

And this is the file I used to look at how Cadence has their Docker setup: https://github.com/cadence-workflow/cadence/blob/master/docker/docker-compose-es-v7.yml

docker-compose/ci-cadence-temporal-dependencies.yml
```diff
@@ -67,43 +67,45 @@ services:
...
+    image: docker.io/bitnami/kafka:3.7
+    hostname: kafka
+    container_name: kafka
     ports:
       - "9092:9092"
     environment:
-      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://kafka:9092
-      KAFKA_LISTENERS: PLAINTEXT://0.0.0.0:9092
-      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+      # KRaft settings
```
lwolczynski, line 87: Same here. Not sure what `KRaft` is

samuel27m: Basically seems to be the way forward so that we no longer need Zookeeper. But mainly, I copied most of the configs from the Cadence repo, so we're doing the same configs as they recommend right now

integ/workflow/wf_ignore_already_started/routers.go
samuel27m: Yes, you can look at the actions in the previous commit in this PR: https://github.com/indeedeng/iwf/pull/548/commits/8194a717a0f5e19e52ead99db4dc0eacbd0cfa35

I believe this was flaky because while the Cadence tests failed(https://github.com/indeedeng/iwf/actions/runs/13464338908/job/37627766325), the All Tests job passed (https://github.com/indeedeng/iwf/actions/runs/13464338906/job/37626732846) and the latter runs the same Cadence tests.

The error was this: https://github.com/indeedeng/iwf/actions/runs/13464338908/job/37627766325#step:4:5143
```
errors.go:6: TestIgnoreAlreadyStartedWorkflowCadence - Test failed with error: strconv.Atoi: parsing "": invalid syntax
```

lwolczynski (approved):

PR #557, opened 2025-04-04
IWF-676: Expand WorkflowInfo to include the workflow type

stevo89519:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #issue_number


service/client/temporal/client.go
```diff
@@ -337,10 +337,13 @@ func (t *temporalClient) DescribeWorkflowExecution(
 
 	memo, err := t.getMemoAndDecryptIfNeeded(resp.GetWorkflowExecutionInfo().GetMemo())
 
+	wfType := resp.GetWorkflowExecutionInfo().GetType().GetName()
```
longquanzheng, line 340: This is  not the type that we should return. This is Temporal workflow type. We need to return IwfWorkflowType which stored as a search attribute. The describeWFExecution API call return it in the SAs list at line 333.

PR #558, opened 2025-04-07
IWF-481: Always use local activity for synchronous update based RPC

### LINKED ISSUE #522: Always use local activities for sync update based RPC
**Context**
Today, whenever an RPC needs locking, we will use sync update of Temporal to process: https://github.com/indeedeng/iwf/blob/20103a8ad81a7a7397610bea8bbd0f584eef3c93/service/api/service.go#L620

Within the sync update, we use an activity to invoke worker RPC API: https://github.com/indeedeng/iwf/blob/20103a8ad81a7a7397610bea8bbd0f584eef3c93/service/interpreter/workflowUpdater.go#L92

The activity will choose to use local activity or regular based on workflow config.

However, the config can be ignored and always use local activity, because there is no side effect of using local activity for this case, but only benefits. The latency will be lower with LA, and saving the storage cost becaues omitting the activity input in LA.

**Goal**
Update the workflowUpdate to always use local activity.

Note:
* Needs a versioning because this is not determinstic change

samuel27m:
### Description
Always use local activity for synchronous update based RPC

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related Issue
Closes #522

### DIFF
--- a/integ/start_with_no_options_test.go
+++ b/integ/start_with_no_options_test.go
@@ -28,10 +28,6 @@ func TestStartWorkflowNoOptionsCadence(t *testing.T) {
 }
 
 func doTestStartWorkflowWithoutStartOptions(t *testing.T, backendType service.BackendType) {
-	if !*cadenceIntegTest {
-		t.Skip()
-	}
-
 	wfHandler := basic.NewHandler()
 	closeFunc1 := startWorkflowWorker(wfHandler, t)
 	defer closeFunc1()
--- a/replayTests/replay_test.go
+++ b/replayTests/replay_test.go
@@ -26,6 +26,8 @@ var jsonHistoryFiles = []string{
 	"v6-search-attributes-optimization-default.json",
 	"v7-no-global-version-search-attribute.json",
 	"v8-yield-on-conditional-complete.json",
+	"v8-activity-for-sync-updates-rpcs.json",
+	"v9-force-local-activity-for-sync-updates-rpcs.json",
 }
 
 func TestTemporalReplay(t *testing.T) {
--- a/service/interpreter/workflowUpdater.go
+++ b/service/interpreter/workflowUpdater.go
@@ -99,7 +99,7 @@ func (u *WorkflowUpdater) handler(
 	var activityOutput interfaces.InvokeRpcActivityOutput
 
 	if u.globalVersioner.IsAfterVersionOfSyncUpdateRPCUseLocalActivity() {
-		err = u.provider.ExecuteActivity(&activityOutput, true, ctx,
+		err = u.provider.ExecuteLocalActivity(&activityOutput, ctx,
 			InvokeWorkerRpc, u.provider.GetBackendType(), &rpcPrep, input)
 	} else {
 		err = u.provider.ExecuteActivity(&activityOutput, u.configer.ShouldOptimizeActivity(), ctx,
--- a/service/interpreter/workflowUpdater.go
+++ b/service/interpreter/workflowUpdater.go
@@ -100,10 +100,10 @@ func (u *WorkflowUpdater) handler(
 
 	if u.globalVersioner.IsAfterVersionOfSyncUpdateRPCUseLocalActivity() {
 		err = u.provider.ExecuteActivity(&activityOutput, true, ctx,
-			InvokeWorkerRpc, u.provider.GetBackendType(), rpcPrep, input)
+			InvokeWorkerRpc, u.provider.GetBackendType(), &rpcPrep, input)
 	} else {
 		err = u.provider.ExecuteActivity(&activityOutput, u.configer.ShouldOptimizeActivity(), ctx,
-			InvokeWorkerRpc, u.provider.GetBackendType(), rpcPrep, input)
+			InvokeWorkerRpc, u.provider.GetBackendType(), &rpcPrep, input)
 	}
 
 	u.persistenceManager.UnlockPersistence(input.SearchAttributesLoadingPolicy, input.DataAttributesLoadingPolicy)
--- a/service/interpreter/workflowUpdater.go
+++ b/service/interpreter/workflowUpdater.go
@@ -97,8 +97,15 @@ func (u *WorkflowUpdater) handler(
 	}
 	ctx = u.provider.WithActivityOptions(ctx, activityOptions)
 	var activityOutput interfaces.InvokeRpcActivityOutput
-	err = u.provider.ExecuteActivity(&activityOutput, u.configer.ShouldOptimizeActivity(), ctx,
-		InvokeWorkerRpc, u.provider.GetBackendType(), rpcPrep, input)
+
+	if u.globalVersioner.IsAfterVersionOfSyncUpdateRPCUseLocalActivity() {
+		err = u.provider.ExecuteActivity(&activityOutput, true, ctx,
+			InvokeWorkerRpc, u.provider.GetBackendType(), rpcPrep, input)
+	} else {
+		err = u.provider.ExecuteActivity(&activityOutput, u.configer.ShouldOptimizeActivity(), ctx,
+			InvokeWorkerRpc, u.provider.GetBackendType(), rpcPrep, input)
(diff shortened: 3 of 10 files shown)


integ/start_with_no_options_test.go
```diff
@@ -28,10 +28,6 @@ func TestStartWorkflowNoOptionsCadence(t *testing.T) {
 }
 
 func doTestStartWorkflowWithoutStartOptions(t *testing.T, backendType service.BackendType) {
-	if !*cadenceIntegTest {
```
samuel27m, line 31: Noticed that this test was being skipped incorrectly. I removed this and the test passes as expected locally, so I assume it was a mistake in the first place.

replayTests/replay_test.go
```diff
@@ -26,6 +26,8 @@ var jsonHistoryFiles = []string{
 	"v6-search-attributes-optimization-default.json",
 	"v7-no-global-version-search-attribute.json",
 	"v8-yield-on-conditional-complete.json",
+	"v8-activity-for-sync-updates-rpcs.json",
+	"v9-force-local-activity-for-sync-updates-rpcs.json",
```
samuel27m, line 30: First time doing replay tests. I added replay tests for before and after the change, is this the correct way of doing it? I thought it'd be safer to check both before and after, rather than just after.

service/interpreter/workflowUpdater.go
```diff
@@ -97,8 +97,15 @@ func (u *WorkflowUpdater) handler(
...
 	var activityOutput interfaces.InvokeRpcActivityOutput
-	err = u.provider.ExecuteActivity(&activityOutput, u.configer.ShouldOptimizeActivity(), ctx,
-		InvokeWorkerRpc, u.provider.GetBackendType(), rpcPrep, input)
+
+	if u.globalVersioner.IsAfterVersionOfSyncUpdateRPCUseLocalActivity() {
+		err = u.provider.ExecuteActivity(&activityOutput, true, ctx,
+			InvokeWorkerRpc, u.provider.GetBackendType(), &rpcPrep, input)
+	} else {
+		err = u.provider.ExecuteActivity(&activityOutput, u.configer.ShouldOptimizeActivity(), ctx,
+			InvokeWorkerRpc, u.provider.GetBackendType(), &rpcPrep, input)
```
samuel27m, line 106: The previous call was:
```
err = u.provider.ExecuteActivity(&activityOutput, u.configer.ShouldOptimizeActivity(), ctx,
		InvokeWorkerRpc, u.provider.GetBackendType(), rpcPrep, input)
```

I changed the `rpcPrep` parameter  to a pointer. Because the `rpcPrep` parameter was not being passed as a pointer before, when the `LocalActivity` tried to execute for `InvokeWorkerRpc` an error always occurred, because the arguments did not match. When that happened, it would fallback to an `Activity`. 
This was happening when users were opting in for `OptimizeActivity` in their config. 

After some investigation, I believe this fell under our radar because our tests mainly check the functionality, but it does not check the Workflow History event types. Because the activities were still being executed (fall back to normal `Activity`) our tests had no reason to know if something was wrong here.

samuel27m: I do have a question for reviewers:

If I change the parameters in the `else` block (i.e. for previous versions), from `rpcPrep` to `&rpcPrep`, would this cause Deterministic errors?
```
err = u.provider.ExecuteActivity(&activityOutput, u.configer.ShouldOptimizeActivity(), ctx,
			InvokeWorkerRpc, u.provider.GetBackendType(), &rpcPrep, input)
```

I'm guessing yes, because if a user opted in for `OptimizeActivity`, this would change the history entry for that execution from `Activity` to `LocalActivity` - but I'm unsure if that's a Deterministic error or not.

longquanzheng: > when the LocalActivity tried to execute for InvokeWorkerRpc an error always occurred, because the arguments did not match

Ohh interesting! 
It that for integ test or replay test? I guess you meant integ test -- If so, do you mean passing a pointer will not get error? that sounds weird. because we are not passing pointer here either: https://github.com/indeedeng/iwf/blob/main/service/interpreter/workflowImpl.go#L621

samuel27m: > It that for integ test or replay test?
For everything - Integration Tests, Replay Tests, and user-built workflows.

> If so, do you mean passing a pointer will not get error? because we are not passing pointer here either: https://github.com/indeedeng/iwf/blob/main/service/interpreter/workflowImpl.go#L621

`workflow.ExecuteActivity` doesn't seem to have a problem when the parameters are not matching their exact declaration, but `workflow.ExecuteLocalActivity` returns an error.

The `InvokeWorkerRpc` method in `activityImpl.go` has the following parameters: 
```
(
	ctx context.Context,
        backendType service.BackendType, 
        rpcPrep *service.PrepareRpcQueryResponse,
	req iwfidl.WorkflowRpcRequest,
)
```
In here, the `rpcPrep` parameter is declared as a pointer.

So IF we don't pass it as a pointer `workflow.ExecuteLocalActivity` will not be able to execute this function as a LocalActivity, and it falls back to normal Activity execution, which doesn't seem to have the same requirements with the parameters. In `workflowProvider.go`:
```
        if optimizeByLocalActivity {
		f := workflow.ExecuteLocalActivity(wfCtx, activity, args...)
		err = f.Get(wfCtx, valuePtr)
		if err != nil {
			f = workflow.ExecuteActivity(wfCtx, activity, args...)
			return f.Get(wfCtx, valuePtr)
		}
		return err
	}
```

In the example you sent, the `StateApiWaitUntil` method in `activityImpl.go` has the following parameters: 
```
(
	ctx context.Context, 
        backendType service.BackendType, 
        input service.StateStartActivityInput, 
        searchAttributes []iwfidl.SearchAttribute,
)
```

And here none of them are pointers, so if we tried executing this with `workflow.ExecuteLocalActivity`, the issue wouldn't occur, because the parameters match perfectly.

longquanzheng: Actually, we don't want to simply use `u.provider.ExecuteActivity(&activityOutput, true ...)`

Because this will use local activity and then fallback to regular. 

The idea is to only use local activity without fallback. There is no value to do fallback in this case. If the invokeRPC HTTP call fails, we just return error to client caller without further retry.

samuel27m: AH! Gotcha. I'll make changes to introduce a new `u.provider.ExecuteLocalActivity` function for this case. Is that what you mean here?

longquanzheng: > The InvokeWorkerRpc method in activityImpl.go has the following parameters:

Ah I see! yeah I think there is a "bug" in Temporal SDK not able to understand pointer param for local activity like regular activity. Good catch!

replayTests/replay_test.go
lwolczynski: We have been adding a test after the change but adding one for the before and after is perfectly fine 👍

service/interpreter/workflowUpdater.go
samuel27m: Made changes now - The workflow provider interface has been updated to include a `ExecuteLocalActivity`, and method `u.provider.ExecuteLocalActivity` method is used for update based RPCs

samuel27m:
The codecov jobs are failing because there is [no test coverage](https://app.codecov.io/gh/indeedeng/iwf/pull/558?src=pr&el=tree&filepath=service/interpreter/cadence/workflowProvider.go#diff-c2VydmljZS9pbnRlcnByZXRlci9jYWRlbmNlL3dvcmtmbG93UHJvdmlkZXIuZ28=) for the Cadence implementation. That is because, at the moment, the Cadence implementation does not support Locking. Therefore, sync update based RPCs cannot be executed. As per [this test](https://github.com/indeedeng/iwf/blob/main/integ/persistence_loading_policy_test.go#L143-L148)

lwolczynski (approved):

Issue #561, opened 2025-04-16
How to configure Temporal Cloud with iWF?


aroysbox:
Hi, 

I was curious how to configure Temporal Cloud with iwf, using the API key that is present in the Temporal cloud instance.

Thanks,
Alex.

longquanzheng:
Hi I am adding this now: https://github.com/indeedeng/iwf/pull/562

PR #562, opened 2025-04-16
Support Temporal API Key

longquanzheng:
### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related Issue
Closes https://github.com/indeedeng/iwf/issues/561


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/indeedeng/iwf
 
-go 1.22
+go 1.23.0
```
longquanzheng, line 3: Looks like it's required by temporal cloud sdk: https://github.com/temporalio/cloud-sdk-go/blob/main/go.mod

config/config_template.yaml
```diff
@@ -9,14 +9,7 @@ api:
...
-#    {{- end }}
-#    {{- if eq $be "temporal" }}
-#    cadence:
-#      hostPort: {{ default .Env.CADENCE_HOST_PORT "127.0.0.1:7833" }}
-#      domain: {{ default .Env.CADENCE_NAMESPACE "default" }}
-#    {{- end }}
\ No newline at end of file
+    hostPort: {{ default .Env.TEMPORAL_HOST_PORT "temporal:7233" }}
+    namespace: {{ default .Env.TEMPORAL_NAMESPACE "default" }}
+    cloudAPIKey: {{ default .Env.TEMPORAL_CLOUD_API_KEY "" }}
```
samuel27m, line 15: What is the `config_template.yaml` file used for?

go.mod
samuel27m: Could we just do 1.23? So that we're targeting all of the 1.23 go releases instead of 1.23.0 only?

cmd/server/iwf/iwf.go
```diff
@@ -110,21 +111,35 @@ func start(c *cli.Context) {
...
+		} else {
+			temporalClient, err = client.Dial(client.Options{
+				HostPort:       temporalConfig.HostPort,
+				Namespace:      temporalConfig.Namespace,
+				MetricsHandler: metricHandler,
+			})
+		}
+
 		if err != nil {
 			rawLog.Fatalf("Unable to connect to Temporal because of error %v", err)
```
samuel27m, line 140: Can we change this error message to be conditional of the connection type we're doing? i.e. 

```
if err != nil {
    if temporalConfig.CloudAPIKey != "" {
        rawLog.Fatalf("Unable to connect to Temporal Cloud (namespace: %s) because of error: %v", temporalConfig.Namespace, err)
    } else {
        rawLog.Fatalf("Unable to connect to Temporal because of error %v", err)
    }
}
```

samuel27m (commented):
There seems to be some sort of conflict between the 2 temporal libs?
```
panic: proto: file "temporal/api/cloud/region/v1/message.proto" is already registered
		previously from: "go.temporal.io/api/cloud/region/v1"
		currently from:  "go.temporal.io/cloud-sdk/api/region/v1"
	See https://protobuf.dev/reference/go/faq#namespace-conflict
```

cmd/server/iwf/iwf.go
longquanzheng: that's a good idea. However, I have reverted the change to use cloud SDK. The current temporal goSDK does support API key out of the box.  And the cloud SDK is still experiemental.

go.mod
```diff
@@ -6,11 +6,11 @@ require (
 	github.com/gin-gonic/gin v1.9.1
 	github.com/google/uuid v1.6.0
 	github.com/prometheus/client_golang v1.12.1
-	github.com/stretchr/testify v1.9.0
+	github.com/stretchr/testify v1.10.0
 	github.com/uber-go/tally/v4 v4.1.1
 	github.com/uber/cadence-idl v0.0.0-20220713235846-fda89e95df1e
 	github.com/urfave/cli v1.22.5
-	go.temporal.io/sdk v1.30.0
+	go.temporal.io/sdk v1.33.1
```
longquanzheng, line 13: Suggested by temporal team to upgrade to 1.33 otherwise API key may need extra code for namespace header
Screenshot 2025-04-17 at 11 14 44 AM

config/config_template.yaml
longquanzheng: Good catch...!
I was following cadence project to use config [template](https://github.com/cadence-workflow/cadence/blob/master/docker/config_template.yaml) , which can be dynamically generating real config like [this](https://github.com/cadence-workflow/cadence/blob/master/docker/start-cadence.sh#L7).

However, I found this approach too cubersome to maintain because there are too many parameters need to be templated. Instead, user can always just mount their own config using volume mouting:
* docker: https://docs.docker.com/engine/storage/volumes/
* K8s https://kubernetes.io/docs/concepts/storage/volumes/

So I am removing the variables now and keep it simple. 

Now the only reason to have config_template.yaml is that the default config value is different from local development (the hostport)

longquanzheng: Nit: I should also rename config_template.yaml to something else. But that will be a minor breaking changes for whoever has used this in production 🤦 

https://github.com/indeedeng/iwf/issues/563

longquanzheng:
Note: no new test for this PR because we don't usally add this for this part. and it's not easy to test with temporal cloud here. 
But I will have an OSS user testing it before merging

lwolczynski (approved):

PR #564, opened 2025-04-22
IWF-786: update libraries to fix bot vulnerability alerts

dlukiantsev:
Updated libraries:
- golang.org/x/net 
- google.golang.org/protobuf
- github.com/gin-gonic/gin 
- golang.org/x/crypto

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related Issue
Closes #issue_number


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/indeedeng/iwf
 
-go 1.22
+go 1.23.0
```
samuel27m, line 3: Could we just do 1.23? So that we're targeting all of the 1.23 go releases instead of 1.23.0 only?

samuel27m (changes requested):

go.mod
samuel27m: I'm not sure myself, just wondering if it's possible at all to do 😄

dlukiantsev: I tried to keep 1.23, but `go mod tidy` is changing it automatically. 

In any case, 1.23.0 is just indicating the minimum go version that is required to build the module, so it will work with Go 1.23.0, 1.23.1, 1.23.2, and so on.

lwolczynski: > In any case, 1.23.0 is just indicating the minimum go version that is required to build the module, so it will work with Go 1.23.0, 1.23.1, 1.23.2, and so on.

I think it is pinning the specific version. You should be able to keep it as `go 1.23`.

lwolczynski (approved):

samuel27m (approved):

lwolczynski (approved):

Issue #520, opened 2024-12-13
Replace stateStartApiSuccess with stateWaitUntilFailed 


longquanzheng:
As new name and also save the few byes

PR #568, opened 2025-05-16
IWF-471: Add stateWaitUntilFailed to CommandResults

samuel27m:
### Description
Adding stateWaitUntilFailed to CommandResults
The server now sets both values: the original StateStartApiSucceeded and the new StateWaitUntilFailed

From here on, we will make changes in the SDKs to use `StateWaitUntilFailed`. We're keeping both fields right now for backwards compatibility

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related issue
Addressing, but not yet closing #520, as there are more follow up MRs to be done in SDK repos


gen/iwfidl/api/openapi.yaml
```diff
@@ -1342,7 +1342,6 @@ components:
       enum:
       - STATE_DECISION_FAILING_WORKFLOW_ERROR_TYPE
       - CLIENT_API_FAILING_WORKFLOW_ERROR_TYPE
-      - STATE_API_FAIL_MAX_OUT_RETRY_ERROR_TYPE
```
samuel27m, line 1345: Unsure why I'm getting this - did we recently remove this error type in the IDL without updating the Server with the new spec?

lwolczynski: Another reason can be that we haven't updated the IDL files here since the removal of `STATE_API_FAIL_MAX_OUT_RETRY_ERROR_TYPE` in IDL repo

samuel27m: Ah, completely missed that. I did that now, it seems there were other changes but not the one we were looking for here. 

It's most likely that we just forgot to update the IDL files like you mentioned. 

I checked the codebase and it is not in use at the moment, so we should be safe here

lwolczynski (approved):

PR #598, opened 2026-01-08
IWF-1153: checking all S3 objects aginst Temporal

ktrops:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #issue_number

### DIFF
--- a/service/interpreter/activityImpl.go
+++ b/service/interpreter/activityImpl.go
@@ -560,54 +560,52 @@ func doLoadFromExternalStorage(ctx context.Context, input *iwfidl.EncodedObject)
 
 func CleanupBlobStore(
 	ctx context.Context, backendType service.BackendType, storeId string,
-) error {
+) (int, error) {
 	store := env.GetBlobStore()
 	provider := interfaces.GetActivityProviderByType(backendType)
 	logger := provider.GetLogger(ctx)
 	logger.Info("CleanupBlobStore started")
-	cfg := env.GetSharedConfig()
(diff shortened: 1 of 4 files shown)


service/interpreter/activityImpl.go
```diff
@@ -560,54 +560,52 @@ func doLoadFromExternalStorage(ctx context.Context, input *iwfidl.EncodedObject)
...
-			} else {
-				if err != nil {
-					logger.Error("CleanupBlobStore failed to delete workflow objects", "workflowPath", workflowPath, "error", err)
-					return err
+					return totalDeleted, err
 				}
+				totalDeleted++
+				logger.Info("CleanupBlobStore deleted workflow objects", "workflowPath", workflowPath)
+			} else if err != nil {
+				logger.Error("CleanupBlobStore failed to describe workflow", "workflowPath", workflowPath, "error", err)
```
dillia23, line 605: we're returning and throwing an error here if describe workflow fails because that means the workflow is likely no longer running?

dillia23:
overall looks good to me! I would approve this if I could!

service/interpreter/activityImpl.go
ktrops: The other errors describe workflow can return besides NotFound error are:
	//  - serviceerror.InvalidArgument
	//  - serviceerror.Internal
	//  - serviceerror.Unavailable

Describe workflow is an expensive call and can take up lots of resources. So if we get any of the above errors, I think we'd want to fix the issue or wait before running it again.

PR #599, opened 2026-03-11
IWF-1174: add logging for S3 & local activity errors

ktrops:
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### Related Issue
Closes #issue_number

### DIFF
--- a/service/interpreter/activityImpl.go
+++ b/service/interpreter/activityImpl.go
@@ -2,6 +2,7 @@ package interpreter
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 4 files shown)


service/interpreter/activityImpl.go
```diff
@@ -59,13 +60,27 @@ func StateApiWaitUntil(
 	if input.Request.StateInput != nil && input.Request.StateInput.ExtStoreId != nil {
 		_, err = loadStateInputFromExternalStorage(ctx, input.Request.StateInput)
 		if err != nil {
+			if activityInfo.IsLocalActivity {
+				reqBytes, _ := json.Marshal(input.Request)
+				logger.Info("StateApiWaitUntil local activity return on error",
```
longquanzheng: should it be Warn or Error?

ktrops: 👋 
Warn would probably be better. Teams are having trouble understanding why local activity fails and most of the time it's failing due to payload size.

something-something-something:
lgtm

something-something-something (approved):

dillia23:
also looks good to me!

service/interpreter/activityImpl.go
longquanzheng: You should push Temporal to fix their history model: https://github.com/temporalio/temporal/issues/4389
https://github.com/temporalio/temporal/issues/5310 so that iWF can get rid of Local Activity

longquanzheng (commented):
There will be tons of logs . It will be better to enable with a config instead.

something-something-something (approved):

PR #600, opened 2026-03-17
Support consuming N messages from channel in one command

### LINKED ISSUE #301: Support get all messages from channel in one command

pradhankukiran:
### Description
Implements server-side support for `AtLeast`/`AtMost` fields on `InterStateChannelCommand` (fields already exist in IDL but were unused). Enables three consumption modes:
  - **Exact N**: `AtLeast=N, AtMost=N` — wait for exactly N, consume exactly N
  - **OneToAll**: `AtLeast=1` (no AtMost) — wait for at least 1, consume all available
  - **ZeroToAll**: `AtLeast=0` (no AtMost) — don't wait, consume all available

  Messages are consumed atomically. Gated behind global version 11 for backward compat with existing workflows.

  **Note:** The `Values` field on `InterStateChannelResult` was added manually to the generated model. A corresponding update to `iwf-idl/iwf.yaml` is needed so `make idl-code-gen` produces the same result.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### Related Issue
Closes #301

### DIFF
--- a/service/interfaces.go
+++ b/service/interfaces.go
@@ -160,6 +160,9 @@ type (
 		CompletedTimerCommands             map[int]InternalTimerStatus   `json:"completedTimerCommands"`
 		CompletedSignalCommands            map[int]*iwfidl.EncodedObject `json:"completedSignalCommands"`
 		CompletedInterStateChannelCommands map[int]*iwfidl.EncodedObject `json:"completedInterStateChannelCommands"`
+		// CompletedInterStateChannelMultiCmds stores multiple messages consumed by a single command
+		// when AtLeast/AtMost are set. Keyed by command index.
+		CompletedInterStateChannelMultiCmds map[int][]*iwfidl.EncodedObject `json:"completedInterStateChannelMultiCmds,omitempty"`
 	}
 
 	StaleSkipTimerSignal struct {
--- a/service/interfaces.go
+++ b/service/interfaces.go
@@ -157,8 +157,10 @@ type (
 	}
 
 	StateExecutionCompletedCommands struct {
-		CompletedTimerCommands             map[int]InternalTimerStatus   `json:"completedTimerCommands"`
-		CompletedSignalCommands            map[int]*iwfidl.EncodedObject `json:"completedSignalCommands"`
+		CompletedTimerCommands  map[int]InternalTimerStatus   `json:"completedTimerCommands"`
+		CompletedSignalCommands map[int]*iwfidl.EncodedObject `json:"completedSignalCommands"`
+		// CompletedInterStateChannelCommands is still used for legacy single-message channel commands.
+		// New AtLeast/AtMost commands use CompletedInterStateChannelMultiCmds.
 		CompletedInterStateChannelCommands map[int]*iwfidl.EncodedObject `json:"completedInterStateChannelCommands"`
 		// CompletedInterStateChannelMultiCmds stores multiple messages consumed by a single command
 		// when AtLeast/AtMost are set. Keyed by command index.
--- a/service/interpreter/continueAsNewer.go
+++ b/service/interpreter/continueAsNewer.go
@@ -172,6 +172,7 @@ func (c *ContinueAsNewer) AddPotentialStateExecutionToResume(
 	commandRequest iwfidl.CommandRequest,
 	completedTimerCommands map[int]service.InternalTimerStatus,
 	completedSignalCommands, completedInterStateChannelCommands map[int]*iwfidl.EncodedObject,
+	completedInterStateChannelMultiCmds map[int][]*iwfidl.EncodedObject,
 ) {
 	c.StateExecutionToResumeMap[stateExecutionId] = service.StateExecutionResumeInfo{
 		StateExecutionId:     stateExecutionId,
@@ -179,9 +180,10 @@ func (c *ContinueAsNewer) AddPotentialStateExecutionToResume(
 		StateExecutionLocals: stateExecLocals,
 		CommandRequest:       commandRequest,
 		StateExecutionCompletedCommands: service.StateExecutionCompletedCommands{
-			CompletedTimerCommands:             completedTimerCommands,
-			CompletedSignalCommands:            completedSignalCommands,
-			CompletedInterStateChannelCommands: completedInterStateChannelCommands,
+			CompletedTimerCommands:              completedTimerCommands,
+			CompletedSignalCommands:             completedSignalCommands,
+			CompletedInterStateChannelCommands:  completedInterStateChannelCommands,
+			CompletedInterStateChannelMultiCmds: completedInterStateChannelMultiCmds,
 		},
 	}
 }
--- a/service/interpreter/workflowImpl.go
+++ b/service/interpreter/workflowImpl.go
@@ -575,6 +575,7 @@ func processStateExecution(
 	completedTimerCmds := map[int]service.InternalTimerStatus{}
 	completedSignalCmds := map[int]*iwfidl.EncodedObject{}
 	completedInterStateChannelCmds := map[int]*iwfidl.EncodedObject{}
+	completedInterStateChannelMultiCmds := map[int][]*iwfidl.EncodedObject{}
 
 	state := stateReq.GetStateMovement()
 	isResumeFromContinueAsNew := stateReq.IsResumeRequest()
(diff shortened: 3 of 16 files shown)


longquanzheng:
thanks for contribution. Can you add test to make sure it works?

pradhankukiran:
@longquanzheng added integration tests covering all 4 consumption modes (ExactN, OneToAll, ZeroToAll, AtMostOnly) across both Temporal and Cadence backends, with and without ContinueAsNew.                                                                                                                                                             
                
Also fixed a bug where AtLeast=0, AtMost=1 would panic on an empty channel, Retrieve() doesn't handle empty channels, so the code now routes through the safe RetrieveUpToN path when atLeast == 0.

service/interpreter/continueAsNewer.go
```diff
@@ -172,16 +172,18 @@ func (c *ContinueAsNewer) AddPotentialStateExecutionToResume(
 	commandRequest iwfidl.CommandRequest,
 	completedTimerCommands map[int]service.InternalTimerStatus,
 	completedSignalCommands, completedInterStateChannelCommands map[int]*iwfidl.EncodedObject,
+	completedInterStateChannelMultiCmds map[int][]*iwfidl.EncodedObject,
```
longquanzheng, line 175: do you think we can just change `completedInterStateChannelCommands` to be `map[int][]*iwfidl.EncodedObject`? 
Or that's a breaking change that will break the determinism?

longquanzheng (changes requested):

service/interpreter/continueAsNewer.go
pradhankukiran: Changing `completedInterStateChannelCommands` to `map[int][]*EncodedObject` would break continue-as-new compatibility. Existing snapshots have object values, not arrays, so old waiting workflows may fail unmarshal/replay. Separate `CompletedInterStateChannelMultiCmds` keeps old snapshot shape intact.

pradhankukiran:
I am adding a follow-up commit. Plain is to  keep `CompletedInterStateChannelCommands` as-is for continue-as-new compatibility, add `InterStateChannelResult.values` through IDL/codegen, validate invalid `atLeast/atMost`, and add focused unit coverage for the new channel/decider paths.
I’ll update generated Go from the IDL locally, but the canonical `iwf-idl` change may need a separate PR/update in the IDL repo.

longquanzheng:
> I am adding a follow-up commit. Plain is to keep `CompletedInterStateChannelCommands` as-is for continue-as-new compatibility, add `InterStateChannelResult.values` through IDL/codegen, validate invalid `atLeast/atMost`, and add focused unit coverage for the new channel/decider paths. I’ll update generated Go from the IDL locally, but the canonical `iwf-idl` change may need a separate PR/update in the IDL repo.

yes, please open a separate PR in idl repo and merge in that first.

service/interpreter/workflowImpl.go
```diff
@@ -1092,6 +1146,55 @@ func getCommandThreadName(prefix string, stateExecId, cmdId string, idx int) str
...
+
+func validateChannelCommandLimits(cmd iwfidl.InterStateChannelCommand) error {
+	if cmd.HasAtLeast() && cmd.GetAtLeast() < 0 {
+		return fmt.Errorf("InterStateChannelCommand atLeast cannot be negative")
+	}
+	if cmd.HasAtMost() && cmd.GetAtMost() < 0 {
+		return fmt.Errorf("InterStateChannelCommand atMost cannot be negative")
+	}
+	if cmd.HasAtLeast() && cmd.HasAtMost() && cmd.GetAtMost() < cmd.GetAtLeast() {
+		return fmt.Errorf("InterStateChannelCommand atMost cannot be less than atLeast")
```
longquanzheng: returning plain error here will cause workflow task running in a loop. we can return application erorr to fail the workflow directly.

service/interfaces.go
```diff
@@ -160,6 +160,9 @@ type (
 		CompletedTimerCommands             map[int]InternalTimerStatus   `json:"completedTimerCommands"`
 		CompletedSignalCommands            map[int]*iwfidl.EncodedObject `json:"completedSignalCommands"`
 		CompletedInterStateChannelCommands map[int]*iwfidl.EncodedObject `json:"completedInterStateChannelCommands"`
```
longquanzheng, line 164: can you leave a comment here? are we deprecating that or it's still being used?

service/interpreter/workflowImpl.go
```diff
@@ -685,6 +689,10 @@ func processStateExecution(
 		stateExecutionLocal = startResponse.GetUpsertStateLocals()
 	}
 
+	if globalVersioner.IsAfterVersionOfChannelConsumeN() {
```
longquanzheng, line 692: can you explain a bit more why this is necssary here?

service/interpreter/workflowImpl.go
```diff
@@ -773,17 +785,38 @@ func processStateExecution(
...
-					// Note that commandReqDoneOrCanceled is needed for two cases:
-					// 1. will be true when trigger type of the commandReq is completed(e.g. AnyCommandCompleted) so we don't need to wait for all commands. Returning the thread to avoid thread leakage.
-					// 2. will be true to cancel the wait for unblocking continueAsNew(continueAsNew will wait for all threads to complete)
-					return received || commandReqDoneOrCanceled
-				})
+				atLeast, atMost := getChannelCommandLimits(cmd, globalVersioner)
 
-				if received {
-					completedInterStateChannelCmds[idx] = interStateChannel.Retrieve(cmd.ChannelName)
+				for {
```
longquanzheng: I don't think we need a for loop? or maybe I missed something(sorry btw I could be wrong, haven't touched the code for a while)

service/interpreter/workflowImpl.go
```diff
@@ -773,17 +785,38 @@ func processStateExecution(
...
+					if atMost > 1 || atLeast == 0 {
+						if values == nil {
+							values = []*iwfidl.EncodedObject{}
+						}
+						completedInterStateChannelMultiCmds[idx] = values
+					} else {
+						if len(values) > 0 {
+							completedInterStateChannelCmds[idx] = values[0]
+						} else {
+							completedInterStateChannelMultiCmds[idx] = values
```
longquanzheng: I feel like it's better not to always backfill the legacy field for single cmd? this means activity input will have duplicate data which cause the history size grow faster.

Since SDK need to change to support this anyway, we should change the SDK to only use the new field only.

For compatibility of old sdk, it will not send the mult-command, server will still need to check if it's single command, then fallback to the old code path and use the single command completed fields.

pradhankukiran:
@longquanzheng Opened the IDL PR here: https://github.com/indeedeng/iwf-idl/pull/91

service/interpreter/workflowImpl.go
```diff
@@ -773,17 +787,40 @@ func processStateExecution(
...
-					return received || commandReqDoneOrCanceled
-				})
-
-				if received {
-					completedInterStateChannelCmds[idx] = interStateChannel.Retrieve(cmd.ChannelName)
+				atLeast, atMost := getChannelCommandLimits(cmd, globalVersioner)
+				multiMessageCommand := isMultiMessageChannelCommand(cmd, globalVersioner)
+
+				// Another command thread may consume from the same channel after this
+				// thread's Await condition is true. Re-check during retrieval; if data
```
longquanzheng: > may consume from the same channel after this thread's Await condition is true

I don't think that's possible in Cadence/Temporal's threading model(cooperative instead of preemtive). 

Are you sure about this? Maybe you can write a test to confirm that behaviro for me. 

Basically, once a thread(goroutine) got return from Await, it will continue to run until next blocking operation. There will not be any interruption between that.

pradhankukiran:
Good point. Removed the retry loop and the comment. Since workflow goroutines are cooperative, once Await returns this thread runs the synchronous retrieval without another command thread interleaving first, so a single consume path is enough.

service/interpreter/workflowImpl.go
longquanzheng: I still don't understand why we have to do this. can you elaborate? maybe with an example? 
thanks

service/interpreter/workflowImpl_test.go
longquanzheng, line 1: we don't usually have test like this

longquanzheng:
> @longquanzheng Opened the IDL PR here: [indeedeng/iwf-idl#91](https://github.com/indeedeng/iwf-idl/pull/91)

oh , you also need to do it for iwf-sdk.yaml https://github.com/indeedeng/iwf-idl/blob/main/iwf-sdk.yaml

PR #606, opened 2026-06-22
IWF-1201: upgrade Temporal Go SDK to v1.44.1

AkilFranklin:
### Description

Upgrades `go.temporal.io/sdk` from v1.30.0 to **v1.44.1** (`go.temporal.io/api` → v1.62.12) and raises the module's `go` directive to **1.24**, which the new SDK requires.

Changes accompanying the dependency bump:

- **Regenerated `service/client/temporal/real_temporal_client_mock.go`** against the v1.44 client interfaces (mockgen source mode). The regenerated mock picks up the new `client.Client` methods (e.g. `CompleteActivityByActivityID`) and no longer emits `MockCloudOperationsClient`. That mock imported `go.temporal.io/api/cloud/cloudservice/v1`, which Temporal removed from `go.temporal.io/api` as of **api v1.41.0** (the Cloud API moved to the separate `go.temporal.io/cloud-sdk` module). The unused cloud mock was what blocked building against any Temporal Go SDK `>= v1.31`.
- **`service/client/temporal/client.go` — encrypted-memo decoding fix.** Newer SDKs apply the configured `DataConverter` (including its `PayloadCodec`) to memos directly ([sdk-go #1045](https://github.com/temporalio/sdk-go/issues/1045)); older SDKs used the default converter. iwf also pre-encrypts memo values, so the stored memo is double-wrapped by the encrypting converter. `getMemoAndDecryptIfNeeded` now decodes encrypted memos **twice through the encrypting converter** instead of *default-then-crypto*. This fixes the `*WithMemoAndEncryption` integration tests, and stays compatible with memos written by older iwf/SDK versions (the codec passes through non-matching payload encodings, so the legacy default-wrapped form still decodes).
- **`service/api/service.go`**: `fmt.Errorf(errMsg)` → `fmt.Errorf("%s", errMsg)`, to satisfy Go 1.24's stricter `printf` vet check.

### ⚠️ Consumer impact

- **Minimum Go version is now 1.24** (was 1.23).
- The Temporal Go SDK and transitive deps (grpc, protobuf, genproto, …) are pulled up via MVS for consumers that adopt this release.
- iwf's interpreter now runs on SDK v1.44.1 — **consumers with in-flight workflows should re-validate replay determinism after upgrading.** Validated against recent production histories: **0 non-determinism** on v1.44.1.
- SDK v1.41 introduced per-client worker heartbeating, which expects Temporal **server >= v1.29.1**.

### Checklist
- [x] Code compiles correctly
- [x] Tests for the changes have been added — existing `*WithMemoAndEncryption` integration tests cover the memo decode change; all pass on v1.44.1
- [x] All tests passing — build, vet, unit tests, and the Temporal integration suite (incl. memo + encryption) pass locally
- [x] **This PR change is backwards-compatible** (no iwf API change; it raises the Go/SDK floor — see Consumer impact)

### Related Issue
Tracked internally as IWF-1201.


something-something-something (approved):
lgtm

service/client/temporal/client.go
```diff
@@ -402,8 +401,13 @@ func (t *temporalClient) getMemoAndDecryptIfNeeded(memo *common.Memo) (map[strin
 	for k, payload := range memo.GetFields() {
 
 		if t.memoEncryption {
-			var encryptedPayload commonpb.Payload
-			err := converter.GetDefaultDataConverter().FromPayload(payload, &encryptedPayload)
+			// Newer Temporal SDKs apply the configured DataConverter (including its
+			// PayloadCodec) to memos (sdk-go #1045), whereas older SDKs used the default
+			// converter. iwf also pre-encrypts the memo value, so the stored memo is
+			// double-wrapped by the encrypting converter. Decode twice through that same
+			// converter: first to recover the inner (pre-encrypted) payload, then the value.
```
inferatus, line 408: This seems a little wonky.  Is this a breaking change?

AkilFranklin: Not a breaking change — `getMemoAndDecryptIfNeeded` is internal and its contract is unchanged; it returns the same decoded values.

It looks odd because the memo is double-wrapped. iwf pre-encrypts the memo value, and newer Temporal SDKs now *also* apply the configured DataConverter (incl. its codec) to memos — that's the sdk-go #1045 fix. Older SDKs used the *default* converter for memos, which is why the previous code did default-decode then crypto-decode. With the new SDK the outer wrapper is the encrypting converter too, so both decode steps go through it.

It's backward-compatible with memos written by older iwf/SDK versions: the codec passes through payloads whose encoding isn't "binary/encrypted", so a legacy default-wrapped memo still decodes via the same two steps. The added unit test (service/client/temporal/memo_test.go) covers both the new and legacy formats.

